PR #1, opened 2022-07-11
Fix some errors, fix some performance warnings, and some simple (but effective) optimizations

toxieainc:


Src/Graphics/Render2D.cpp
```diff
@@ -333,23 +333,13 @@ static inline void DrawTileLine(uint32_t *line, int pixelOffset, uint16_t tile,
     uint32_t pattern = vram[patternOffset + patternLine];
     for (int p = 7; p >= 0; p--)
     {
-      if (!clip || (clip && pixelOffset >= 0 && pixelOffset < 496))
+      if (!clip || (/*pixelOffset >= 0 &&*/ (unsigned int)pixelOffset < 496u))
```
trzy, line 336: The test for pixelOffset >= 0 is required or you will end up writing below index 0 of the surface buffer. Although the buffers are 512 lines wide, rendering begins at offset 0.

Src/Graphics/Render2D.cpp
```diff
@@ -361,23 +351,13 @@ static inline void DrawTileLine(uint32_t *line, int pixelOffset, uint16_t tile,
       uint32_t pattern = vram[patternOffset + patternLine + i];
       for (int p = 3; p >= 0; p--)
       {
-        if (!clip || (clip && pixelOffset >= 0 && pixelOffset < 496))
+        if (!clip || (/*pixelOffset >= 0 &&*/ (unsigned int)pixelOffset < 496u))
```
trzy, line 354: Likewise here. The pixelOffset check is required. I would prefer the check remain for readability as 2D rendering is not a significant performance sink for now. Another option would be to shift the pixel buffers by 8. But I would need to check more carefully to ensure the buffers are laid out as I recall. It appears they are 512*384 at first glance. This solution would need to be clearly documented, however.

Src/Util/ByteSwap.cpp
```diff
@@ -1,19 +1,41 @@
...
 namespace Util
 {
-  void FlipEndian16(uint8_t *buffer, size_t size)
+  void FlipEndian16(uint8_t * const buffer, const size_t size)
   {
+#ifdef _MSC_VER
+    uint16_t * const buffer16 = (uint16_t*)buffer;
+    for (size_t i = 0; i < size/2; i++)
+      buffer16[i] = _byteswap_ushort(buffer16[i]);
+#elif 1 // assume GCC
```
trzy, line 14: Is there a way to detect GCC? I think there is a GNUC or some such macro defined. This would be preferred than leaving a pointless #else that is never used.

Src/CPU/68K/Musashi/m68kdasm.c
```diff
@@ -251,8 +251,8 @@ static const char* g_cpcc[64] =
 {/* 000    001    010    011    100    101    110    111 */
 	  "f",  "eq", "ogt", "oge", "olt", "ole", "ogl",  "or", /* 000 */
 	 "un", "ueq", "ugt", "uge", "ult", "ule",  "ne",   "t", /* 001 */
-	 "sf", "seq",  "gt",  "ge",  "lt",  "le",  "gl"  "gle", /* 010 */
-  "ngle", "ngl", "nle", "nlt", "nge", "ngt", "sne",  "st", /* 011 */
+	 "sf", "seq",  "gt",  "ge",  "lt",  "le",  "gl", "gle", /* 010 */
```
trzy, line 254: Oooh missing comma, good catch! Does this affect MAME too?

Src/GameLoader.cpp
```diff
@@ -193,11 +193,11 @@ static void PopulateGameInfo(Game *game, const Util::Config::Node &game_node)
   game->name = game_node["name"].ValueAs<std::string>();
   game->parent = game_node["parent"].ValueAsDefault<std::string>(std::string());
   game->title = game_node["identity/title"].ValueAsDefault<std::string>("Unknown");
-  game->version = game_node["identity/version"].ValueAsDefault<std::string>("");
+  game->version = game_node["identity/version"].ValueAsDefault<std::string>(std::string());
```
trzy, line 196: Any reason for this in particular? I find the former a bit easier on the eyes.

Src/Network/TCPSend.cpp
```diff
@@ -60,7 +60,7 @@ bool TCPSend::Send(const void * data, int length)
 
 	DPRINTF("Sending %i bytes\n", length);
 
-	int sent = 0;
+	int sent;
```
trzy, line 63: Not a fan of unintialized vars. Why not:

```
int sent = SDLNet_TCP_Send(m_socket, &length, sizeof(int));
```

Src/GameLoader.cpp
```diff
@@ -159,7 +159,7 @@ bool GameLoader::File::operator==(const File &rhs) const
 
 GameLoader::Region::ptr_t GameLoader::Region::Create(const GameLoader &loader, const Util::Config::Node &region_node)
 {
-  if (GameLoader::MissingAttrib(loader, region_node, "name") | MissingAttrib(loader, region_node, "stride") | GameLoader::MissingAttrib(loader, region_node, "chunk_size"))
+  if (GameLoader::MissingAttrib(loader, region_node, "name") || MissingAttrib(loader, region_node, "stride") | GameLoader::MissingAttrib(loader, region_node, "chunk_size"))
```
trzy, line 162: The | rather than || is intentional. We do *not* want to short-circuit because these functions print useful error messages. With short-circuiting, we would not get a comprehensive error readout. Because this at first glance seems non-obvious, please add a comment to at least one of these (but maybe all of these). E.g.:

```
Non-short-circuiting | is intentional so that all error messages are printed.
```

Src/GameLoader.cpp
```diff
@@ -135,7 +135,7 @@ bool GameLoader::MissingAttrib(const GameLoader &loader, const Util::Config::Nod
 
 GameLoader::File::ptr_t GameLoader::File::Create(const GameLoader &loader, const Util::Config::Node &file_node)
 {
-  if (GameLoader::MissingAttrib(loader, file_node, "name") | GameLoader::MissingAttrib(loader, file_node, "offset"))
+  if (GameLoader::MissingAttrib(loader, file_node, "name") || GameLoader::MissingAttrib(loader, file_node, "offset"))
```
trzy, line 138: | is correct, || is wrong.

Src/Inputs/Inputs.cpp
```diff
@@ -332,7 +332,8 @@ void CInputs::PrintHeader(const char *fmt, ...)
 	va_end(vl);
 
 	puts(header);
-	for (size_t i = 0; i < strlen(header); i++)
+	const size_t hs = strlen(header);
```
trzy, line 335: Non-performance critical, prefer readable form over extra line of code.

trzy (changes requested):
Please see the individually addressed items and fix. Thank you!

Src/Graphics/Render2D.cpp
toxieainc: This one is just an equivalent optimization. As its cast to unsigned, it includes the >=0 check implicitly, thus i left it there, but commented out.

Src/Graphics/Render2D.cpp
toxieainc: In my profiling this inner loop actually was very noticable. Not the most important, but always under the Top 5.

Src/Util/ByteSwap.cpp
toxieainc: I just copied this from other parts of the code base, but will fix it nevertheless.

Src/CPU/68K/Musashi/m68kdasm.c
toxieainc: No, MAME already had this fixed a long time ago. My next PR will have some more things that are missing/fishy in the old m68k core.

Src/Inputs/Inputs.cpp
toxieainc: No problem. I just addressed the IMHO sane performance warnings. Will revert.

Src/GameLoader.cpp
toxieainc: Not really, just a performance warning.

toxieainc:
I think i addressed all now, please check again.

Src/Graphics/Render2D.cpp
toxieainc: Done

trzy:
Looks great now!

PR #2, opened 2022-07-12
Games.xml: Align ROM names with MAME.

ToBul:
Minor and inconsequential change.
Makes manual ROM building 'slightly' easier when using latest MAME ROM set (currently 0.245).


trzy:
Thanks for this and the other patch. I will review it then this weekend.

trzy (approved):

PR #17, opened 2022-08-19
Small optimizations and fix pointer aliasing

toxieainc:
Harmless math warning fixes/optimizations, no functional change (apart from being faster as no double<->float casts).
Change map->unordered_map in CNew3D::CacheModel (as no order needed there).
Avoid general pointer aliasing problems (using bit_cast (C++20 if available), or the recommended/compiler-optimized-away memcpy) to be more future proof.

### DIFF
--- a/Src/Graphics/New3D/R3DFloat.cpp
+++ b/Src/Graphics/New3D/R3DFloat.cpp
@@ -35,5 +35,5 @@ UINT32 R3DFloat::Convert16BitProFloat(UINT32 a1)
 
 float R3DFloat::ToFloat(UINT32 a1)
 {
-	return Util::UintAsFloat(a1);
+	return Util::Uint32AsFloat(a1);
 }
--- a/Src/Graphics/New3D/R3DFloat.cpp
+++ b/Src/Graphics/New3D/R3DFloat.cpp
@@ -1,5 +1,6 @@
 #include "Types.h"
 #include "R3DFloat.h"
+#include "Util\GenericValue.h"
 
 float R3DFloat::GetFloat16(UINT16 f)
 {
@@ -13,17 +14,16 @@ float R3DFloat::GetFloat32(UINT32 f)
 
 UINT32 R3DFloat::ConvertProFloat(UINT32 a1)
 {
-	int exponent = (a1 & 0x7E000000) >> 25;
+	UINT32 exponent = (a1 & 0x7E000000) >> 25;
 
 	if (exponent <= 31) {	// positive
 		exponent += 127;
 	}
 	else {					// negative exponent
-		exponent -= 64;
-		exponent += 127;
+		exponent += 127 - 64;
 	}
 
-	int mantissa = (a1 & 0x1FFFFFF) >> 2;
+	UINT32 mantissa = (a1 & 0x1FFFFFF) >> 2;
 
 	return (a1 & 0x80000000) | (exponent << 23) | mantissa;
 }
@@ -35,5 +35,5 @@ UINT32 R3DFloat::Convert16BitProFloat(UINT32 a1)
 
 float R3DFloat::ToFloat(UINT32 a1)
 {
-	return *(float*)(&a1);
+	return uint_as_float(a1);
 }
--- a/Src/Graphics/New3D/R3DFloat.cpp
+++ b/Src/Graphics/New3D/R3DFloat.cpp
@@ -1,6 +1,6 @@
 #include "Types.h"
 #include "R3DFloat.h"
-#include "Util/GenericValue.h"
+#include "Util/BitCast.h"
 
 float R3DFloat::GetFloat16(UINT16 f)
 {
@@ -35,5 +35,5 @@ UINT32 R3DFloat::Convert16BitProFloat(UINT32 a1)
 
 float R3DFloat::ToFloat(UINT32 a1)
 {
-	return uint_as_float(a1);
+	return Util::UintAsFloat(a1);
 }
--- a/Src/Graphics/New3D/R3DFloat.cpp
+++ b/Src/Graphics/New3D/R3DFloat.cpp
@@ -1,6 +1,6 @@
 #include "Types.h"
 #include "R3DFloat.h"
-#include "Util\GenericValue.h"
+#include "Util/GenericValue.h"
 
 float R3DFloat::GetFloat16(UINT16 f)
 {
--- a/Src/Util/BitCast.h
+++ b/Src/Util/BitCast.h
@@ -1,0 +1,61 @@
+#ifndef INCLUDED_UTIL_BITCAST_H
+#define INCLUDED_UTIL_BITCAST_H
+
+#include <stdexcept>
+#include <cstring>
+
+namespace Util
+{
+#if __cplusplus >= 202002L
+#include <bit>
+#define FloatAsInt(x) std::bit_cast<int>(x)
+#define IntAsFloat(x) std::bit_cast<float>(x)
+#define UintAsFloat(x) std::bit_cast<float>(x)
+#elif 1
+template <class Dest, class Source>
+inline Dest bit_cast(Source const& source) {
+   static_assert(sizeof(Dest) == sizeof(Source), "size of destination and source objects must be equal");
+   static_assert(std::is_trivially_copyable<Dest>::value, "destination type must be trivially copyable.");
+   static_assert(std::is_trivially_copyable<Source>::value, "source type must be trivially copyable");
+
+   Dest dest;
+   std::memcpy(&dest, &source, sizeof(dest));
+   return dest;
+}
+#define FloatAsInt(x) bit_cast<int,float>(x)
+#define IntAsFloat(x) bit_cast<float,int>(x)
+#define UintAsFloat(x) bit_cast<float,unsigned int>(x)
+#else
+inline int FloatAsInt(const float x)
+{
+   union {
+      float f;
+      int i;
+   } uc;
+   uc.f = x;
+   return uc.i;
+}
+
+inline float IntAsFloat(const int i)
+{
+   union {
+      int i;
+      float f;
+   } iaf;
+   iaf.i = i;
+   return iaf.f;
+}
+
+inline float UintAsFloat(const unsigned int i)
+{
+   union {
+      unsigned int u;
+      float f;
+   } iaf;
+   iaf.u = i;
+   return iaf.f;
+}
+#endif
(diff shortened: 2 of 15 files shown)


dukeeeey:
Unordered map isn't automatically faster. The implementation is usually some form of hash table so will likely allocate a good chunk of memory. For smaller tables just a regular map can be a lot more efficient. 

I don't see the point of bit cast. I'm sure it'll compile to identical code. It just obfuscate exactly what the code is doing for me.

toxieainc:
I would have totally agreed on both 10 or 15 years ago.

The mem overhead of unordered maps should be neglibile here, simply as its overall just a tiny fraction of the overall mem usage of Supermodel. Performance wise it is in the vast majority of cases superior with all 'common' compiler versions from the last years.

As for the bit casts: The problem is that compilers (especially gcc and also more and more clang) get extremely aggressive nowadays, and everything that is essentially undefined behavior can and will be exploited. So unless one specifies special compile options, strict aliasing is always enabled on the higher optimization levels.

See e.g. https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8
and for GCC: 'The -fstrict-aliasing option is enabled at levels -O2, -O3, -Os.'

For MSVC this is usually not a problem (at the moment!), but i've seen gcc 'break' such code already in our own production code.

dukeeeey:
For maps it depends on the use case. If you create an unordered map inside a function call and it only lives for the duration of the function call there is a good chance you'll hit the kernel/os to allocate more memory from the heap. This is the expensive part and where the slowness comes from. Small maps are much less likely to have this problem as the memory can come from already allocated memory from a memory pool managed by malloc. Of course this answer is highly implementation dependent but is generally true. So you can't just blindly swap out map for a hash table version and expect it to go faster. But if you can profile the code and it's a hot spot and a hash table version is faster then of course it's a no brainer to swap them out.

As for the casting code. If I am  being pedantic you made the c++ function into a macro function which is generally a horrible idea. If you don't know exactly what you are doing with parenthesis you can get entirely the wrong result. That is one of the weaknesses of using macros.

toxieainc:
I agree for the general case. But in this case the macro is directly wiring the param through to each function call, which IMHO makes the sideeffects 0, as it looks like a function and actually behaves like the called function, too.

toxieainc:
I also just did some benchmarking (windows/VS2019) regarding unordered_map vs map. Timing was done using the high resolution timer/perf counter, accumulating over 10000s of CacheModel calls. Overall its no large difference, but unordered_map is a bit faster across multiple tested games.

trzy:
I haven’t yet looked at this in detail (traveling for another week or so) and don’t have a strong opinion about the map data structure.

Otherwise, as a general principle, I think readability should be prioritized above performance (unless we are going for a *big* perf win), with deference given to the person who will be spending most time maintaining the code (almost always the person who wrote it). 

I do think that it is worthwhile to clean up the various compiler warnings being spewed.

So let’s see if we can come to a compromise here.

Src/Graphics/New3D/R3DFloat.cpp
```diff
@@ -1,5 +1,6 @@
 #include "Types.h"
 #include "R3DFloat.h"
+#include "Util\GenericValue.h"
```
trzy, line 3: Shouldn’t we use forward slash / for paths?

toxieainc: True, thanks, corrected.

toxieainc:
As for readability vs perf: Absolutely, but the bit casts are there for conformance with cpp standards. As mentioned otherwise, it may(!) happen that compilers exploit this undefined behavior (i.e. potentially resulting in hard to track down bugs).
I'm also not keen on using these, but as compilers are really aggressive nowadays.. .. ..

dukeeeey:
Normal casts are part of the core language and will be there forever. Bit_cast can be done at compile time for say constant expressions. But this doesn't really help us in supermodel

toxieainc:
These will be there forever, yes. And i also used these with great joy ;) for years and years.

BUT compilers nowadays can (and will) exploit the spec, see the link above (https://gist.github.com/shafik/848ae25ee209f698763cffee272a58f8).
So either one has to use some 'official' workarounds (bit_cast or memcpy) or has to disable strict aliasing (e.g. -fno-strict-aliasing for gcc, like parts of Linux does).

toxieainc:
From https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#Type-punning:
'In particular, an object of one type is assumed never to reside at the same address as an object of a different type, unless the types are almost the same. For example, an unsigned int can alias an int, but not a void* or a double.'
So for the places i changed, this is exactly the case. So i don't like this fact, too, but nowadays it is like it is.  :/

dukeeeey:
Okay, makes sense.

ToBul:
@toxieainc
Sorry for butting in but I tried your branch; to get it to build in gcc I had to remove the 'std::' scope from memcpy.

trzy:
Hey sorry this has been stuck in PR hell. I'll be back home on Friday and can look at this on Saturday. I think it's probably safe to merge.

Src/Util/GenericValue.h
```diff
@@ -20,6 +20,57 @@
...
+{
+   union {
+      int i;
+      float f;
+   } iaf;
+   iaf.i = i;
+   return iaf.f;
+}
+
+inline float uint_as_float(const unsigned int i)
```
trzy, line 63: How does this differ from a reinterpret_cast? Why not use one?

trzy (commented):
I think I'm about ready to merge this in. Sorry for the delay. Just had a question about this use of unions rather than a more straightforward reinterpret cast.

Src/Util/GenericValue.h
toxieainc: Its a can of worms and totally stupid/unnecessary c++ mess, really. :(
So far i found that only bit_cast and memcpy are 'accepted' and defined-behavior variants to do such specific casts.
The additional union code i only left in for another possible way (which only GCC explicitly specs as 'fine' though, as one of its language extensions).

If you look at https://en.cppreference.com/w/cpp/language/reinterpret_cast it also avoids to really get into details for everything except pointer/reference stuff, and thus it also seems to be regarded by the community as UB for doing these specific casts.  :/

I also verified that memcpy and bit_cast both end up in optimized away code via godbolt. (e.g. https://godbolt.org/z/M713r79fT  &  https://godbolt.org/z/3K7b3dbvP )

trzy: I think for the task of converting from a uint8_t array, reinterpret_cast actually does work as intended. And certainly bit_cast should. reinterpret_cast<> is supposed to work for converting objects that are represented as std::byte, char, or unsigned char (and I assume uint8_t basically falls into this category).

Are you sure a union is well defined here? Like couldn't, in theory, this use of a union be UB?

dukeeeey: My understanding of the spec in c this is allowed but in c++ it's undefined behaviour. But basically every compiler will support this.

In c++ byte pointer to another type isn't undefined behaviour. 

I do agree that it's a good idea to get away from undefined behaviour. But maybe there are better ways we can fix this in the code such as returning void* and just casting to type we want. Rather than returning int* then recasting. I don't think our existing code is problematic though because we are never writing one type and reading back as another with the exception of my real3dpro float code. That could potentially trigger undefined behaviour I believe. It should be fixed.

dukeeeey: From the spec

It's undefined behavior to read from the member of the union that wasn't most recently written.
https://en.cppreference.com/w/cpp/language/union

Goes on to say
Many compilers implement, as a non-standard language extension, the ability to read inactive members of a union.

toxieainc: Absolutely agreed, thats also why the union code is not used (there is a `#if 1` so that always the memcpy path is used if `bit_cast`/C++20 is not already available to the compiler yet).

Which places you guys think i should _not_ use these new `(u)int_as_float` or `float_as_(u)int`s then?

toxieainc: ..or actually, as i only changed 3 places: Which `ToFloat` locations should i change in addition (i.e. to not use the new functions)?

trzy: Looks like these functions are used in Legacy3D.cpp and R3DFloat.cpp. I didn't notice the #elif 1. I see now what's going on -- I had skimmed that and missed it and was confused. I think these functions are fine then. But here's what I recommend:

1. Remove them from GenericValue.h. Doesn't seem like they belong here. Maybe their own header file?
2. They should be in the Util namespace. 
3. Naming convention should match the rest of Supermodel. E.g., camel case. Util::ToFloat() or something like that.

toxieainc: Okay, i just commited your requests

Src/Util/BitCast.h
```diff
@@ -0,0 +1,61 @@
...
+   static_assert(std::is_trivially_copyable<Source>::value, "source type must be trivially copyable");
+
+   Dest dest;
+   std::memcpy(&dest, &source, sizeof(dest));
+   return dest;
+}
+#define FloatAsInt(x) bit_cast<int,float>(x)
+#define IntAsFloat(x) bit_cast<float,int>(x)
+#define UintAsFloat(x) bit_cast<float,unsigned int>(x)
+#else
```
trzy, line 28: This is never executed if I understand correctly. Let's remove it altogether.

Src/Util/BitCast.h
```diff
@@ -0,0 +1,61 @@
...
+#define INCLUDED_UTIL_BITCAST_H
+
+#include <stdexcept>
+#include <cstring>
+
+namespace Util
+{
+#if __cplusplus >= 202002L
+#include <bit>
+#define FloatAsInt(x) std::bit_cast<int>(x)
```
trzy, line 11: Should we be more specific here and replace Int and Uint with Int32 and UInt32 (using either INT32/UINT32 or int32_t/uint32_t)?

trzy (changes requested):

Src/Util/BitCast.h
trzy: That is, the #else clause with the union stuff.

toxieainc: Done

Src/Util/BitCast.h
toxieainc: Done

trzy (approved):
Looks great!

PR #18, opened 2022-08-20
Reinstate a Star Wars hack

ToBul:
Restores missing textures in the tunnel run near the end of the game.
As per the old hack, it would need to be removed if JTAG was ever emulated.

### DIFF
--- a/Src/Graphics/New3D/New3D.cpp
+++ b/Src/Graphics/New3D/New3D.cpp
@@ -1670,7 +1670,12 @@ void CNew3D::SetSunClamp(bool enable)
 
 void CNew3D::SetSignedShade(bool enable)
 {
-	m_shadeIsSigned = enable;
+	if (m_gameName == "swtrilgy" || m_gameName == "swtrilgya") {
+		m_shadeIsSigned = false;
+	}
+	else {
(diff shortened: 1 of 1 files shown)


ToBul:
Not sure if Supermodel wants to go down the path of dirty hacks, it's just an idea really.
It's something that pops up on the forums every now and again and this is the code I add to my builds.
I'm no dev or coder, so I won't be the slightest bit offended if rejected or changed.

trzy:
I wonder if we can expose this as a Games.xml attribute? I think we already have a couple dubious ones there, including one to override the bridge controller type for a particular game (which itself is odd because the bridge controller IC depends on the board stepping).

We also had something to override certain PCI codes.

ToBul:
I think Supermodel users will be happy and appreciative with any solution you can come up with.
As I said, I'm no coder, so I'm afraid I can't be of any real help.
As a work around people can use -legacy3d or just hack their own builds, so there is no rush.
If you want to close this, that's cool with me.

dukeeeey:
We should just have a go at doing jtag properly. I started writing something but in between power cuts and the magic virus never had a chance to finish it.

ToBul:
Ok, cool. I'll close this from my side.

PR #26, opened 2022-10-05
remove 2 dupe function calls, where ProcessLos should actually be rat…

toxieainc:
…her costly

I think both calls are not really intended, but maybe something else was, at some point?

### DIFF
--- a/Src/Graphics/New3D/New3D.cpp
+++ b/Src/Graphics/New3D/New3D.cpp
@@ -369,11 +369,9 @@ void CNew3D::RenderFrame(void)
 			SetRenderStates();
 
 			m_r3dShader.DiscardAlpha(true);						// discard all translucent pixels in opaque pass
-			m_r3dFrameBuffers.SetFBO(Layer::colour);
 			bool hasOverlay = RenderScene(pri, renderOverlay, Layer::colour);
 
 			if (!renderOverlay && ProcessLos(pri)) {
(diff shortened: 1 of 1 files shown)


toxieainc:
Just checked the history, the first one is a leftover from changing the loops around.
The 2nd one i cannot explain why it is there (its like that since the initial commit), but at least it reads the exact same value back, which should be costly i'd assume, unless the driver does some magical caching under the hood.

dukeeeey:
Yeah that extra call probably shouldn't be there. The line of sight stuff is only used by scud if I recall. It works enough to emulate scud correctly but not 100% sure the logic was correct. There is some LOS flag in the polyheader which maybe does something, I'm not sure.  The sky in scud is marked with that flag. But other games set it to But never use the line of sight stuff.

But back to the thing the extra call should be removed. But I don't think it'll have any performance impact. Calling glreadpixels once for the depth is expensive because you create a sync point in the driver. Basically the driver has to flush the pipeline to return your value. But doing it a second time is basically free.

toxieainc:
Okay. So is there anything that should happen if the call fails or succeeds in addition?

dukeeeey:
Currently no. I think the intention was probably to have some other logic in there. I think it's safe to remove.

toxieainc:
Good, ready to go then IMHO.

trzy:
Thanks!

PR #27, opened 2022-10-05
Games.xml: magtruck (partially) redumped

ToBul:
Bad dump patch no longer needed. Region change patch still works.

MAME commit,
https://github.com/mamedev/mame/commit/b24695eda73a7206a75f432eddbd88508e8c62d4

Only 2 ROMS have actually changed. MAME have oddly changed the version to 'Export', which seems wrong, as it defaults to Japan without the region enable patch.

### DIFF
--- a/Config/Games.xml
+++ b/Config/Games.xml
@@ -1237,15 +1237,13 @@
     </hardware>
     <roms>
       <patches>
-	    <!-- Fixing bad ROM dump which causes attract mode to stop rendering -->
-        <patch region="crom" bits="32" offset="0x091A34" value="0x917F0068" />
         <!-- Enable Region change (remove the patch to return to Japan region) -->
         <patch region="crom" bits="32" offset="0x14B25C" value="0x38600001" />
       </patches>
       <region name="crom" stride="8" chunk_size="2" byte_swap="true">
-        <file offset="0" name="epr-21434.20" crc32="0xE028D7CA" />
+        <file offset="0" name="epr-21434.20" crc32="0x18F8626A" />
         <file offset="2" name="epr-21436.19" crc32="0x22BCBCA3" />
-        <file offset="4" name="epr-21433.18" crc32="0x60AA9D76" />
+        <file offset="4" name="epr-21433.18" crc32="0xB3B124FA" />
         <file offset="6" name="epr-21435.17" crc32="0x9B169446" />
       </region>
       <region name="banked_crom" stride="8" chunk_size="2" byte_swap="true">


trzy:
This is great. I have these ROMs already (in fact it was me who requested they be redumped). But I have one question: should we apply this as-is or is it worth creating a second alt ROM set definition that includes the old, corrupt ROMs plus patch? I'm not sure how most people find their ROMs and concerned many users will still be finding the old ones. 

On the other hand, that's a lot of trouble and maybe not worth it. We can make it a policy to always use the latest and greatest ROM sets. Thoughts?

dukeeeey:
I'd support the old rom + patch too for the simple reason it would be less support requests lol

LodanZark:
Since supermodel doesn't supply a DAT file to rebuild\verify the ROM set, syncing the ROMs detail with the latest mame releases would be a better solution.
Probably better to keep the old ROMs until the mame 0.249 release, then update with redumped files.

ToBul:
@trzy, I had the same thought process as you, concluding it's best to sync with MAME. It should help avoid confusion and extra work.
Maybe make a post in the 'Automated Builds' thread back at the forum when a ROM set changes, so people know to update the set in question or stick with their old Games.xml.
As @LodanZark has suggested, the MAME magtruck commit missed the 0.248 merge anyway, so this PR is now essentially a month early. Best wait till MAME 0.249 if the goal is to stay in step with their ROM sets.

Having said all that, if you can get the Games.xml to catch all scenarios that would be a bonus. You wouldn't have to worry about being perfectly aligned with MAME and it would save Ian shaking his head :). Not sure of the best way to do that.

As always, if you want to change or reject this PR to meet your decision, that's cool.

normchomsky:
After playing around with MAME roms from 0.236 I can say that I have had no issue or missing roms (other than the optional rom files) with new Git versions of Supermodel. I have tested multiple sets with multiple versions of the newer git builds. TBH I thought you guys had already synced up with MAME.

LodanZark:
Not every single ROM it's updated with newer mame releases, it just happens magtruck got redumped and added in the current release (mame 0.249).

ToBul:
The model3 ROM set will of course be very static, and dumps will only get increasingly rare.

normchomsky:
Cool I get it. So going forward from here are you guys updating the Games.xml to run both the new dump and the old dump of this gamelike ToBul suggested above? Or is it going to be 'new version = new 0.249 dump or keep old Games.xml'?

trzy:
I've pushed a commit that renames the old ROM set with patch to mgtrkbad for people still using it and added a child set, magtruck, with the updated ROMs and no patch. Thanks for submitting this -- it was just easier to make the modification myself. Please withdraw.

PR #28, opened 2022-10-11
small optimizations and small precision improvements to the quad and …

toxieainc:
…tri shaders

rather harmless and also IMHO does not harm readability, etc, maybe even rather the contrary


trzy:
Will try to look at this and the other PR this weekend!

trzy (commented):
Quite a lot of changes here. Some I can understand but others I am not sure of. Has Ian confirmed that this doesn't break anything?

dukeeeey:
> 

I haven't tried the code but it looks fine. Personally I'd leave the code the same for LinearTexLocations in the quad shader. The micro optimisations at least for me obfuscate the meaning of what the code is doing

trzy:
toxieainc: possible to make Ian's suggested fix so we can merge this in?

toxieainc:
Yup, will do.

toxieainc:
I reverted everything in LinearTexLocations as requested

toxieainc:
Will check what is leftover from this commit after @dukeeeey did the GL revamp

PR #35, opened 2022-11-02
Add uint declaration to line 1903 of m68kcpu.h

TheBrinkOfTomorrow:
Fix long-standing compiler error on macOS.

Submitting per my comments in "Supermodel macOS Binaries for latest SVN release..." forum thread...

### DIFF
--- a/Src/CPU/68K/Musashi/m68kcpu.h
+++ b/Src/CPU/68K/Musashi/m68kcpu.h
@@ -1899,7 +1899,10 @@ static inline void m68ki_exception_interrupt(uint int_level)
 		return;
 
 	/* Acknowledge the interrupt */
-	vector = m68ki_int_ack(int_level);
+#if defined(__APPLE__)
+    uint m68ki_int_ack(uint);
+#endif
+    vector = m68ki_int_ack(int_level);
 
 	/* Get the interrupt vector */
 	if(vector == M68K_INT_ACK_AUTOVECTOR)


TheBrinkOfTomorrow:
Submitting per my comments in "Supermodel macOS Binaries for latest SVN release..."  forum thread...

trzy:
I'm not comfortable with this solution. Trying to fool the compiler by supplying a fake prototype to a function that doesn't actually exist seems like a really brittle hack. What exactly is the error on MacOS? The macros should expand without problem.

dukeeeey:
What's the actual error message it spits out?
uint is defined in the header file so I don't get what it's complaining about

dukeeeey:
I think the error is this

supermodel\Src\CPU\68K\Musashi\m68kcpu.h(1902,11): warning C4013: 'M68KIRQCallback' undefined; assuming extern returning int

Maybe it is assumes int type it won't work with unsigned int?

This should have been fixed ages ago >_< but it worked so i never cared enough to fix it

dukeeeey:
I think the correct solution to this problem is to define 

int M68KIRQCallback(int nIRQ);

On line 215 of m68kconf.h

I haven't 100% got my head around how this code is working because it's dynamically creating .c files and pulling in various headers. But I think this is enough to fix it. Maybe Bart or someone can check :)

TheBrinkOfTomorrow:
The actual error is:

trzy:
Interesting. I'll have a look again tonight. On MacOS, is it building with clang?

trzy:
Okaaaay. I see what's going on.

So, Musashi has a few ways of specifying the interrupt callback. One can also specify none at all and use autovectoring (where the vector is chosen automatically, which is how most 68K systems are set up, including Model 3). We provide a callback mainly for the debugger to hook into.

m68kconf.h is intended to be edited to set various compile-time configuration options in the Musashi 68K core (which as Ian points out is partly compile-time generated). One of those options is M68K_EMULATE_INT_ACK. If it is set to OPT_SPECIFY_HANDLER, then M68K_INT_ACK_CALLBACK(a) should be set to a function to call directly. Otherwise, if you set it to OPT_ON, then a function pointer that has to be set at run-time in the M68K context will be called. Here's the code in m68kcpu.h that determines what happens:

```
/* Enable or disable callback functions */
#if M68K_EMULATE_INT_ACK
	#if M68K_EMULATE_INT_ACK == OPT_SPECIFY_HANDLER
		#define m68ki_int_ack(A) M68K_INT_ACK_CALLBACK(A)
	#else
		#define m68ki_int_ack(A) CALLBACK_INT_ACK(A)
	#endif
#else
	/* Default action is to used autovector mode, which is most common */
	#define m68ki_int_ack(A) M68K_INT_ACK_AUTOVECTOR
#endif /* M68K_EMULATE_INT_ACK */
```

CALLBACK_INT_ACK(A) resolves to m68ki_cpu.int_ack_callback(A) (that's a function pointer). And we actually set this function pointer, needlessly, on line 302 of 68K.cpp:

```
bool M68KInit(void)
{
	m68k_init();
	m68k_set_cpu_type(M68K_CPU_TYPE_68000);
	m68k_set_int_ack_callback(M68KIRQCallback);
	s_Bus = NULL;
#ifdef SUPERMODEL_DEBUGGER
	s_Debug = NULL;
#endif // SUPERMODEL_DEBUGGER
	DebugLog("Initialized 68K\n");
	return OKAY;
}
```

So we have two choices:

1) Use OPT_ON and then we don't need to define M68K_INT_ACK_CALLBACK at all. This will result in an indirect function call when firing interrupts, which is no big deal because it happens a few times per frame max.

2) Right above that line declare: extern int M68KIRQCallback(int nIRQ); and indicate that it must be consistent with the definition in 68K.h. We shouldn't just include 68K.h because it includes Musashi headers.

I'm actually kind of leaning toward (1). And if we do this we need to change a couple other lines. We also have a per-instruction hook when debugging is enabled, and I incorrectly set M68KReset as the reset instruction callback but luckily left that option as OPT_OFF. Checkout my branch: bart/m68k-conf-fix. What do you all think?

I tested this with and without ENABLE_DEBUGGER (and I suggest you test both on MacOS as well and make sure you can switch into the 68K debugger).

TheBrinkOfTomorrow:
> Interesting. I'll have a look again tonight. On MacOS, is it building with clang?

Yes, clang 14.  On macOS 13.0, here's the x86-64 version info:
```
% cc --version
Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: x86_64-apple-darwin22.1.0
Thread model: posix
```

TheBrinkOfTomorrow:
> Okaaaay. I see what's going on.
> 
> I tested this with and without ENABLE_DEBUGGER (and I suggest you test both on MacOS as well and make sure you can switch into the 68K debugger).

I built and tested your branch on x86 and a M1 SOC, and it works as expected (minimal testing with Scud Race Plus).

However, I get linker errors when I try making with "ENABLE_DEBUGGER=1"  ( on both x86 and Apple Silicon).  
I'm guilty of not having read the Supermodel debugging documentation, so not sure what the expected outcome is...
Here's the end of the make output (in x86, but M1 was very similar).
```
% make -f ./Makefiles/Makefile.OSX NET_BOARD=1 ENABLE_DEBUGGER=1 
.
.
.
1092 warnings generated.
--------------------------------------------------------------------------------
Linking Supermodel     : bin/supermodel
--------------------------------------------------------------------------------
ld: warning: option -s is obsolete and being ignored
Undefined symbols for architecture x86_64:
  "M68KGetRegister(SM68KCtx*, unsigned int)", referenced from:
      Debugger::CMusashi68KDebug::GetReg(Debugger::CCPUDebug*, unsigned int) in Musashi68KDebug.o
      Debugger::CMusashi68KDebug::GetSP() in Musashi68KDebug.o
  "M68KSetRegister(SM68KCtx*, unsigned int, unsigned int)", referenced from:
      Debugger::CMusashi68KDebug::SetReg(Debugger::CCPUDebug*, unsigned int, unsigned int) in Musashi68KDebug.o
  "ppc_attach_debugger(Debugger::CPPCDebug*)", referenced from:
      Debugger::CPPCDebug::AttachToCPU() in PPCDebug.o
  "ppc_detach_debugger()", referenced from:
      Debugger::CPPCDebug::~CPPCDebug() in PPCDebug.o
      Debugger::CPPCDebug::~CPPCDebug() in PPCDebug.o
      Debugger::CPPCDebug::~CPPCDebug() in PPCDebug.o
      Debugger::CPPCDebug::DetachFromCPU() in PPCDebug.o
  "CZ80::AttachDebugger(Debugger::CZ80Debug*)", referenced from:
      Debugger::CZ80Debug::AttachToCPU() in Z80Debug.o
  "CZ80::DetachDebugger()", referenced from:
      Debugger::CZ80Debug::~CZ80Debug() in Z80Debug.o
      Debugger::CZ80Debug::~CZ80Debug() in Z80Debug.o
      Debugger::CZ80Debug::~CZ80Debug() in Z80Debug.o
      Debugger::CZ80Debug::DetachFromCPU() in Z80Debug.o
  "CZ80::GetReg8(unsigned int)", referenced from:
      Debugger::CZ80Debug::ReadReg8(Debugger::CCPUDebug*, unsigned int) in Z80Debug.o
      Debugger::CZ80Debug::GetHandlerAddr(Debugger::CInterrupt*, unsigned int&) in Z80Debug.o
  "CZ80::SetReg8(unsigned int, unsigned char)", referenced from:
      Debugger::CZ80Debug::WriteReg8(Debugger::CCPUDebug*, unsigned int, unsigned char) in Z80Debug.o
  "CZ80::GetReg16(unsigned int)", referenced from:
      Debugger::CZ80Debug::ReadReg16(Debugger::CCPUDebug*, unsigned int) in Z80Debug.o
      Debugger::CZ80Debug::GetReturnAddr(unsigned int, unsigned int, unsigned int&) in Z80Debug.o
  "CZ80::SetReg16(unsigned int, unsigned short)", referenced from:
      Debugger::CZ80Debug::WriteReg16(Debugger::CCPUDebug*, unsigned int, unsigned short) in Z80Debug.o
      Debugger::CZ80Debug::UpdatePC(unsigned int) in Z80Debug.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [bin/supermodel] Error 1
```

trzy:
Did you make clean before making again?

TheBrinkOfTomorrow:
> Did you make clean before making again?

Yep, I cloned your branch directly into a totally separate location, and  I did a "make clean" between the standard and DEBUG builds...

trzy:
Wait looks like you edited your comment and I didn't see that. So the debug build ... builds?

If so, the way to test it is to just run a game and then at some point press Alt-B which should freeze it and you should see in the terminal that the debugger console is active. In that console, you should be able to type:

sc 1

To switch to the Sound68K, for example, and then step a few instructions with:

n

Press enter to issue each command (it's a line oriented input). 

Then to resume execution:

c

If that seems to work then I think we are golden! I don't actually know if anyone has ever tested the debugger on MacOS before.

TheBrinkOfTomorrow:
> Wait looks like you edited your comment and I didn't see that. So the debug build ... builds?

Yeah, I realized I hadn't cleaned as soon as I posted, and edited it as quickly as I could.  Sorry...

> If so, the way to test it is to just run a game and then at some point press Alt-B which should freeze it and you should see in the terminal that the debugger console is active. In that console, you should be able to type:

That all worked.  Some output using the M1 Mac mini:
```
MainPPC @ $00031F60: Execution halted.
Analysing MainPPC...
MainPPC Jmp00031F60  lbz     r4,-$7FC9(r13) > Snd68K @ $6005BA: Execution halted.
sc 1
Analysing Snd68K...
Snd68K IntAUTO3     movem.l D0-D1/D3/A2,-(A7) > n
Snd68K $6005BE      move.b  $100405,D0 > n
Snd68K $6005C4      tst.b   $000C3F > n
Snd68K $6005C8      beq     Jmp6005D0 > n
Snd68K Jmp6005D0    tst.b   D0 > 
Snd68K $6005D2      bmi     Jmp6005EE > 
Snd68K Jmp6005EE    cmpi.b  #$FF,D0 > 
Snd68K $6005F2      beq     Jmp600672 > c
```
> If that seems to work then I think we are golden! I don't actually know if anyone has ever tested the debugger on MacOS before.
Yes, it seems to be fine; and on an ARMv8.x CPU too...  ;-)

trzy:
Thanks a lot! I merged my commit in and am closing this out.

Issue #22, opened 2022-09-16
Noticeable "clicking" sound in music


MDashK:
Using "**Supermodel_0.3a-git-0eef09b_Win64**" version.

Playing Daytona USA 2 Power Edition (also noticeable on Sega Rally 2) there is a random clicking in the bgm sound, I'm assuming it's related to the MPEG decoding.

Here is a ~1m sample where you can clearly listen to the "clicking" multiple times. This seems to happen randomly and sporadicly, but more noticeable in drum beats.

Advised to use earphones or headphones to be more noticeable.

https://mega.nz/file/Mok2zJoJ#OJEvYkO8Nb9kgO3d4btWktHsqocc-V_yRslEO6aVEH8

UPD: Forgot to mention that this did not happen on previous builds like r862. Same PC. Same setup. Nothing changed.

toxieainc:
Maybe due to this? https://github.com/trzy/Supermodel/commit/94d8a1b22d32ddc46a40f8f72fc9b34e55deb4c9

MDashK:
Checking the code you mentioned reminded me of the SoundVolume and MusicVolume options in the config file.
So I changed my settings from:

SoundVolume = 150
MusicVolume = 200

... To ...

SoundVolume = 75
MusicVolume = 125

... And the clicking is gone.

I initially tried with 50/100 values, the clicking was gone but I found it too low, so I upped them to 100/150 and the clicking came back, despite very rarely. 75/125 seems to be the sweet spot.

dukeeeey:
I actually didn't know you could set the volume higher than 100%. The clicking you hear is because the waveform is being flattened in the peaks. This will happen in most audio software if you do the same especially with only 16 bit audio.

trzy:
I didn't know much about audio programming when I wrote the sound mixing and resampling code. That should be performed by SDL automatically using its own internal functions. The linear interpolation for resampling is probably naive and I'll bet audiophiles could perceive a difference between the 16KHz MPEG audio being played at its native 16KHz, rescaled the "right way" (whatever that is) to 44.1KHz, and rescaled my way to 44.1KHz.

[...]

PR #37, opened 2022-11-03
change all INT16 mixing/intermediate clamping to floats

toxieainc:
also fixes 3 bugs:
1) mpeg right channel volume attenuation was always using the left channel volume instead
2) too high MusicVolume setting was not clamped to 0..200
3) too high SoundVolume setting was not clamped to 0..200

partially addresses issue #22

### DIFF
--- a/Src/Inputs/Input.cpp
+++ b/Src/Inputs/Input.cpp
@@ -91,7 +91,7 @@ const char* CInput::GetInputGroup()
 		case Game::INPUT_JOYSTICK1:       // Fall through to below
 		case Game::INPUT_JOYSTICK2:       return "4-Way Joysticks";
 		case Game::INPUT_FIGHTING:        return "Fighting Game Buttons";
-		case Game::INPUT_SPIKEOUT:		 return "Spikeout Buttons";
+		case Game::INPUT_SPIKEOUT:        return "Spikeout Buttons";
 		case Game::INPUT_SOCCER:          return "Virtua Striker Buttons";
 		case Game::INPUT_VEHICLE:         return "Racing Game Steering Controls";
 		case Game::INPUT_SHIFT4:          return "Racing Game Gear 4-Way Shift";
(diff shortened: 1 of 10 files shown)


trzy:
Question: doesn't float reduce precision? As I recall, mixing was currently being done with fixed point arithmetic. I think I was using 24.8.

toxieainc:
In principle this would be true. If one would always keep the 24.8 data around and would never go to 16bits inbetween (but if you look at the code, there are multiple stages, each time one has 16bit data at the end). Also all the ops always do truncation, also when doing the volume attenuation precomputations, etc. So this way one also introduces an implicit systematic error.
Obviously this is all not earth shattering, but as the float only code should even be faster nowadays, i think getting these 1-2 bits/sample improvement and no systematic error is better.  ;)
To do everything 'correctly' one would also need to add some dithering on the final float->INT16 step, but i think this is not really worth it for our use case.

toxieainc:
Oh, and i forgot to mention: If the 18bit DAC path is ever triggered by some games (didn't check), then my change will also keep that data unclamped until the final stage.
Although i also must admit that i don't really understand where these additional 2bits are compensated for. If a game would trigger the same volume range as another one in the 16bit DAC path, then the 18bit one will always be 4x the scale in the end when mixing. Before, it was just clamped (=bad), but now it could just be the full 18bits.
Do you have a clue how this is supposed to be handled? And which games trigger the 18bit DAC path?

toxieainc:
Addon: Seems like at least the majority of all the games triggers the 18bit DAC path.
So this means that ALL these were potentially cut off/clamped on loud sounds! So the 2nd PR will then definetly fix this. This will change the loudness though of everything coming out of SCSP, but its the correct thing to do!

trzy:
There's an 18-bit DAC on the SCSP? That's interesting and a great catch! This PR looks good! Thanks!

toxieainc:
Thanks! As for the 18 bit DAC: Just did some tests on all kind of games, and seems that even when waving the 18bit data through (which in theory could be 4x the value range), then only Daytona triggers some clamping, and only for a handful of samples now and then.

dukeeeey:
I haven't really looked at the audio code much .. But 18bit DAC? I didn't even know this was a thing. Why does it exist? :)

toxieainc:
No clue for the specific case here. But at least it seems like the board can work at >=18bit fidelity. And then there are 2 DAC types, a 18bit and a 16bit one. The 16bit one obviously will slash this additional fidelity, while the 18bit one keeps it when going to analogue output(s).
But as mentioned above, seems like fortunately it doesn't really matter for SuperModel, as only Daytona2 actually (rarely) seems to trigger it.
Otherwise, i haven't checked what happens if one would change the volume in the menu of other games (at least for Daytona2 its already at max by default).

trzy:
> I haven't really looked at the audio code much .. But 18bit DAC? I didn't even know this was a thing. Why does it exist? :)

How else do you expect to "feel the heartbeat of my machine, through this tight seat" if not with 18-bit fidelity??

Issue #38, opened 2022-11-04
Compiling with NET_BOARD on macOS


GregoireLD:
Hi

I successfully build and ran Supermodel on my M1 Arm Mac.
Yet, when I tried to build it again with NET_BOARD enabled, I got an error referring to <SDL2_net/SDL_net.h> not being found.

both SDL_net and SDL2 are installed using brew.

Am I missing something here ? let me know if I can run or check anything you need.

TheBrinkOfTomorrow:
It works fine for me on M1 and x86 Macs with v2.01 of the SDL_net framework
 (built from source from: https://libsdl.org/projects/old/SDL_net/#source )

What version does homebrew include?
[I  purposely haven't tried the homebrew version of SDL_net as I don't want it to interfere with my already working fine installations].

If it helps, I've previously shared a precompiled versions of supermodel and the SDL_net framework (for M1 and x86) on the Supermodel forums (search for thread "Supermodel macOS Binaries for latest SVN release (r865)" ).

dukeeeey:
Question for you mac guys. Are you using the new3d engine? What is the performance like?

TheBrinkOfTomorrow:
> Question for you mac guys. Are you using the new3d engine? What is the performance like?

I just replied to your post on the forum.  :-)

I'll paste it here....

_new3d engine has never worked on macOS / OS X as far back as I've been building/running it.
I've assumed that it's just that Apple's [long since deprecated] OpenGL implementation is missing necessary features required for the new3d engine.
[I'm sure the M1 GPU is more than sufficient to run it well..]

We discussed it here: [viewtopic.php?f=5&t=1911#p18044](http://www.supermodel3.com/Forum/viewtopic.php?f=5&t=1911#p18044)
There's a screen shot and console output in that post... of what the new3d engine looks like.

I just built cloned and built the latest code fromGithub, and here's the info output and the error I get with "-new3d - quad-rendering" - and the graphics are mostly corrupted/missing... same as my previous post... there's no difference on an x86 machine versus Apple Silicon._

```
% ./supermodel-git-20221104-m1-net   -print-gl-info 
Supermodel: A Sega Model 3 Arcade Emulator (Version 0.3a-WIP)
Copyright 2003-2022 by The Supermodel Team
OpenGL information:

  Vendor                   : Apple
  Renderer                 : Apple M1
  Version                  : 2.1 Metal - 83
  Shading Language Version : 1.20
[...]

PR #39, opened 2022-11-05
Add automatic detection for Brew-installed SDL2 and SDL_net

GregoireLD:
Here is a PR regarding my compiling problem on Mac, namely detecting and using Brew-installed SDL2 and SDL_net libraries instead of manually installing them.

refers to : [issue#38](https://github.com/trzy/Supermodel/issues/38)


GregoireLD:
Will add it back when ready for primetime

PR #41, opened 2022-11-08
add undefd codepath for correct use of the 18bit DAC path (which seem…

toxieainc:
…s to be triggered for all games)

the volume correction to bring the data back into a valid range is not really needed in practice though, only Daytona2 seems to need it, and also only extremely rarely, so lets just live with a tiny bit of clamping for that game then

while at it, make some formatting similar to MAME, and add one comment regarding a most likely wrong recent MAME change

### DIFF
--- a/Src/OSD/SDL/Audio.cpp
+++ b/Src/OSD/SDL/Audio.cpp
@@ -142,30 +142,6 @@ void SetAudioType(Game::AudioTypes type)
     AudioType = type;
 }
 
-static INT16 AddAndClampINT16(INT32 x, INT32 y)
-{
-    INT32 sum = x + y;
-    if (sum > INT16_MAX) {
-        sum = INT16_MAX;
-    }
-    if (sum < INT16_MIN) {
-        sum = INT16_MIN;
-    }
-    return (INT16)sum;
-}
-
-static INT16 MixINT16(INT32 x, INT32 y)
-{
-    INT32 sum = (x + y)>>1;
-    if (sum > INT16_MAX) {
-        sum = INT16_MAX;
-    }
-    if (sum < INT16_MIN) {
-        sum = INT16_MIN;
-    }
-    return (INT16)sum;
-}
-
 static INT16 MixINT16(float x, float y)
 {
     INT32 sum = (INT32)((x + y)*0.5f); //!! dither
@@ -185,7 +161,7 @@ static float MixFloat(float x, float y)
 
 static INT16 ClampINT16(float x)
 {
-    INT32 xi = (INT32)x;
+    INT32 xi = (INT32)x; //!! dither
     if (xi > INT16_MAX) {
         xi = INT16_MAX;
     }
--- a/Src/Sound/SCSP.cpp
+++ b/Src/Sound/SCSP.cpp
@@ -85,6 +85,7 @@ bool legacySound; // For LegacySound (SCSP DSP) config option.
 
 #define MAX_SCSP	2
 
+//#define CORRECT_FOR_18BIT_DAC
 
 // These globals control the operation of the SCSP, they are no longer extern and are set through SCSP_SetBuffers(). --Bart
 static double SoundClock; // Originally titled SysFPS; seems to be for the sound CPU.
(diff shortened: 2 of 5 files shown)


toxieainc:
This adds what was discussed in https://github.com/trzy/Supermodel/pull/37 already, but undefd. Otherwise it will just cause confusion for users as they would have to change their volumes.
And as there is now, after the other sound change, rarely any clamping happening at all at default volume levels (100,100).. .. ..

Src/OSD/SDL/Audio.cpp
```diff
@@ -185,7 +161,7 @@ static float MixFloat(float x, float y)
 
 static INT16 ClampINT16(float x)
 {
-    INT32 xi = (INT32)x;
+    INT32 xi = (INT32)x; //!! dither
```
trzy, line 164: What do you mean here by dither?

Src/Sound/SCSP.cpp
```diff
@@ -719,10 +720,10 @@ bool SCSP_Init(const Util::Config::Node &config, int n)
 		else
 			SDL=0.0;
 
-		if(iSDL==0x6)
-			int a=1;
-		if(iTL==0x3a)
-			int a=1;
+		//if(iSDL==0x6)
```
trzy, line 723: What were these lines doing and if they really aren't needed, why not simply delete them?

Src/Sound/SCSP.cpp
```diff
@@ -1667,30 +1671,53 @@ void SCSP_DoMasterSamples(int nsamples)
 		{
 			smpfl = ICLIP18(smpfl);
 			smpfr = ICLIP18(smpfr);
+
+#ifdef CORRECT_FOR_18BIT_DAC
```
trzy, line 1675: Understand the concern about user volumes needing to be changed but if this pathway is really correct, why not go ahead and enable it? Alternatively we could save that for a subsequent PR?

Src/Sound/SCSP.cpp
```diff
@@ -1498,6 +1497,11 @@ signed int inline SCSP_UpdateSlot(_SLOT *slot)
 			UINT16 Enc = ((TL(slot)) << 0x0) | (0x7 << 0xd);
 			*RBUFDST = (sample * LPANTABLE[Enc]) >> (SHIFT + 1);
 		}
+		else
+		{
```
trzy, line 1501: Was this a fix ported from MAME?

Src/OSD/SDL/Audio.cpp
toxieainc: That ideally one should add some noise on top to dither, to get rid of systematic quantization errors.

Src/Sound/SCSP.cpp
toxieainc: I have no clue, i guess somebody used these for debugging? Should i just remove these?

Src/Sound/SCSP.cpp
toxieainc: After all my experiments, it fortunately does not matter in practice to leave this undefd.
Otherwise people would have to fiddle with their per-game settings, and basically for no gain in the end.

trzy: Does the 18-bit DAC make everything quieter? It doesn't all get normalized in the end?

toxieainc: Thats exactly what the whole problem is about. Previously, the 18bit data was simply just taken as if it was 16bit data (so the 18bit and 16bit DAC paths were doing exactly the same at the final stage).
So 4x too 'loud'. Fortunately, only Daytona, and only VERY rarely exceeds that 16bit data range (it seems).

toxieainc: But if we now correct this (by /4-ing), then previous user settings will be 4x too low.

trzy:
Okay, let's merge this in. We probably should at some point just enable this, user settings be damned :) Better that it suddenly gets quieter than 4x louder :D

dukeeeey:
Is it actually 4x louder on the original h/w

toxieainc:
I have no clue. One would have to pull/measure this directly from the soundchip/DAC (pre-amp) i fear.

dukeeeey:
Indeed. Older systems with 8 bit audio weren't super quiet, they just sounded worse. I think 16bit cd audio is supposed to give around 96db of range which is probably way louder than any arcade cabinet.

toxieainc:
If this ever gets revisited, one should IMHO also rethink what is happening with the balance of master and slave within `SCSP_DoMasterSamples()`. Cause IMHO this balancing should happen at a different point of the mixing, cause this way it seems like it could actually affect the emulation behavior itself (although i haven't understood all of the details yet, it just seems to be a bit fishy).

PR #42, opened 2022-11-13
[Makefile.Win32] $MSYSTEM check never fails.

ToBul:
'make clean' from Win command shell doesn't fully clean, condition always true.


trzy:
Great catch!

PR #44, opened 2022-11-14
Keep Supermodel files (config, nvram, saves, etc.) in predictable paths

casasfernando:
Keep Supermodel files (config, nvram, saves, etc.) in predictable paths when running on Linux.
Be default it will check for Config directory presence in current directory, then it will check in $HOME/.supermodel and finally it will check in $HOME/.config/supermodel.
It will also automatically create any missing directories for NVRAM, saves, screenshots, etc. If $HOME/.config/supermodel is used then all directories except Config will be created in $HOME/.local/share/supermodel.

### DIFF
--- a/Src/OSD/FileSystemPath.h
+++ b/Src/OSD/FileSystemPath.h
@@ -32,8 +32,9 @@
 
 namespace FileSystemPath
 {
+    enum fsPathType { Analysis, Config, Log, NVRAM, Saves, Screenshots }; // Filesystem path types
     bool PathExists(std::string fileSystemPath); // Checks if a directory exists (returns true if exists, false if it doesn't)
-    std::string GetPath(std::string pathType);  // Generates a path to be used by Supermodel files
+    std::string GetPath(fsPathType pathType);  // Generates a path to be used by Supermodel files
 }
 
 
--- a/Src/OSD/FileSystemPath.h
+++ b/Src/OSD/FileSystemPath.h
@@ -32,9 +32,9 @@
 
 namespace FileSystemPath
 {
-    enum fsPathType { Analysis, Config, Log, NVRAM, Saves, Screenshots }; // Filesystem path types
+    enum PathType { Analysis, Config, Log, NVRAM, Saves, Screenshots }; // Filesystem path types
     bool PathExists(std::string fileSystemPath); // Checks if a directory exists (returns true if exists, false if it doesn't)
-    std::string GetPath(fsPathType pathType);  // Generates a path to be used by Supermodel files
+    std::string GetPath(PathType pathType);  // Generates a path to be used by Supermodel files
 }
 
 
--- a/Src/OSD/FileSystemPath.h
+++ b/Src/OSD/FileSystemPath.h
@@ -1,0 +1,40 @@
+/**
+ ** Supermodel
+ ** A Sega Model 3 Arcade Emulator.
+ ** Copyright 2003-2022 The Supermodel Team
+ **
+ ** This file is part of Supermodel.
+ **
+ ** Supermodel is free software: you can redistribute it and/or modify it under
+ ** the terms of the GNU General Public License as published by the Free
+ ** Software Foundation, either version 3 of the License, or (at your option)
+ ** any later version.
+ **
+ ** Supermodel is distributed in the hope that it will be useful, but WITHOUT
+ ** ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ ** FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+ ** more details.
+ **
+ ** You should have received a copy of the GNU General Public License along
+ ** with Supermodel.  If not, see <http://www.gnu.org/licenses/>.
+ **/
+
+/*
+ * FileSystemPaths.h
+ * 
+ * Header file for OS-dependent Supermodel files locations.
+ */
+
+#ifndef INCLUDED_FILESYSTEMPATH_H
+#define INCLUDED_FILESYSTEMPATH_H
+
+#include <string>
+
+namespace FileSystemPath
+{
+    bool PathExists(std::string fileSystemPath); // Checks if a directory exists (returns true if exists, false if it doesn't)
+    std::string GetPath(std::string pathType);  // Generates a path to be used by Supermodel files
+}
+
+
+#endif // INCLUDED_FILESYSTEMPATH_H
--- a/Src/OSD/SDL/Main.cpp
+++ b/Src/OSD/SDL/Main.cpp
@@ -471,15 +471,24 @@ static void SaveFrameBuffer(const std::string& file)
 void Screenshot()
 {
     // Make a screenshot
-    char file[128];
-    std::string info = "Screenshot created: ";
     time_t now = std::time(nullptr);
     tm* ltm = std::localtime(&now);
+    std::string file = Util::Format() << FileSystemPath::GetPath(FileSystemPath::Screenshots)
+        << "Screenshot_"
+        << std::setfill('0') << std::setw(4) << (1900 + ltm->tm_year)
+        << '-'
+        << std::setw(2) << (1 + ltm->tm_mon)
+        << '-'
+        << std::setw(2) << ltm->tm_mday
+        << "_("
+        << std::setw(2) << ltm->tm_hour
+        << '-'
+        << std::setw(2) << ltm->tm_min
+        << '-'
+        << std::setw(2) << ltm->tm_sec
+        << ").bmp";
 
-    sprintf(file, "%sScreenshot %.4d-%.2d-%.2d (%.2d-%.2d-%.2d).bmp", FileSystemPath::GetPath(FileSystemPath::Screenshots).c_str(), 1900 + ltm->tm_year, 1 + ltm->tm_mon, ltm->tm_mday, ltm->tm_hour, ltm->tm_min, ltm->tm_sec);
-
-    info += file;
-    puts(info.c_str());
+    std::cout << "Screenshot created: " << file << std::endl;
     SaveFrameBuffer(file);
 }
 
--- a/Src/OSD/SDL/Main.cpp
+++ b/Src/OSD/SDL/Main.cpp
@@ -66,6 +66,7 @@
 #include "Util/Format.h"
 #include "Util/NewConfig.h"
 #include "Util/ConfigBuilders.h"
+#include "OSD/FileSystemPath.h"
 #include "GameLoader.h"
 #include "SDLInputSystem.h"
 #include "SDLIncludes.h"
@@ -475,7 +476,7 @@ void Screenshot()
     time_t now = std::time(nullptr);
     tm* ltm = std::localtime(&now);
 
-    sprintf(file, "Screenshot %.4d-%.2d-%.2d (%.2d-%.2d-%.2d).bmp", 1900 + ltm->tm_year, 1 + ltm->tm_mon, ltm->tm_mday, ltm->tm_hour, ltm->tm_min, ltm->tm_sec);
+    sprintf(file, "%sScreenshot %.4d-%.2d-%.2d (%.2d-%.2d-%.2d).bmp", FileSystemPath::GetPath("Screenshots").c_str(), 1900 + ltm->tm_year, 1 + ltm->tm_mon, ltm->tm_mday, ltm->tm_hour, ltm->tm_min, ltm->tm_sec);
 
     info += file;
     puts(info.c_str());
@@ -549,7 +550,7 @@ static void TestPolygonHeaderBits(IEmulator *Emu)
       if ((unknownPolyBits[idx] & mask))
       {
         Emu->RenderFrame();
-        std::string file = Util::Format() << "Analysis/" << GetFileBaseName(s_gfxStatePath) << "." << "poly" << "." << idx << "_" << Util::Hex(mask) << ".bmp";
+        std::string file = Util::Format() << s_analysisPath << GetFileBaseName(s_gfxStatePath) << "." << "poly" << "." << idx << "_" << Util::Hex(mask) << ".bmp";
         SaveFrameBuffer(file);
       }
     }
@@ -565,7 +566,7 @@ static void TestPolygonHeaderBits(IEmulator *Emu)
       if ((unknownCullingNodeBits[idx] & mask))
       {
         Emu->RenderFrame();
-        std::string file = Util::Format() << "Analysis/" << GetFileBaseName(s_gfxStatePath) << "." << "culling" << "." << idx << "_" << Util::Hex(mask) << ".bmp";
+        std::string file = Util::Format() << s_analysisPath << GetFileBaseName(s_gfxStatePath) << "." << "culling" << "." << idx << "_" << Util::Hex(mask) << ".bmp";
         SaveFrameBuffer(file);
       }
     }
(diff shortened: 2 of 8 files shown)


Src/OSD/FileSystemPath.h
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+#ifndef INCLUDED_FILESYSTEMPATH_H
+#define INCLUDED_FILESYSTEMPATH_H
+
+#include <string>
+
+namespace FileSystemPath
+{
+    enum fsPathType { Analysis, Config, Log, NVRAM, Saves, Screenshots }; // Filesystem path types
```
trzy, line 35: fsPathType -> FSPathType or just PathType (consistent with current enum naming convention).

Src/OSD/SDL/Main.cpp
```diff
@@ -475,7 +476,7 @@ void Screenshot()
     time_t now = std::time(nullptr);
     tm* ltm = std::localtime(&now);
 
-    sprintf(file, "Screenshot %.4d-%.2d-%.2d (%.2d-%.2d-%.2d).bmp", 1900 + ltm->tm_year, 1 + ltm->tm_mon, ltm->tm_mday, ltm->tm_hour, ltm->tm_min, ltm->tm_sec);
+    sprintf(file, "%sScreenshot %.4d-%.2d-%.2d (%.2d-%.2d-%.2d).bmp", FileSystemPath::GetPath(FileSystemPath::Screenshots).c_str(), 1900 + ltm->tm_year, 1 + ltm->tm_mon, ltm->tm_mday, ltm->tm_hour, ltm->tm_min, ltm->tm_sec);
```
trzy, line 479: I think this should be converted to a Util::Format() string builder. We currently lack format specifiers for numbers and dates but it should be possible. I think STL now has some sort of date formatters that can be used to assist with building these strings, too?

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+            case Log:
+                strPathType = "Log";
+                break;
+            case NVRAM:
+                strPathType = "NVRAM";
+                break;
+            case Saves:
+                strPathType = "Saves";
+                break;
+            case Screenshots:
```
trzy, line 73: I would add a default case. And I think that case should print an error and bring down Supermodel. We don't have a great way of doing this right now but I think you could probably throw a fatal exception to kill the application. The reason for this is to guard against programmer errors. For example, if someone adds an enum case and forgets to update the switch statements. Some languages will guard against this but C++ doesn't do this.

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+                strPathType = "Screenshots";
+                break;
+        }
+
+        // Get user's HOME directory
+		if (pwd)
```
trzy, line 79: Some sort of indentation issues here.

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+    {
+        std::string finalPath = "";
+        std::string homePath = "";
+        std::string strPathType = "";
+		struct passwd* pwd = getpwuid(getuid());
+
+        // Resolve pathType to string for later use
+        switch (pathType)
+        {
+            case Analysis:
```
trzy, line 58: Prefer case statements to be aligned with the braces of the switch statement. That is, remove one indent level.

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    }
+
+    // Generates a path to be used by Supermodel files
+    std::string GetPath(fsPathType pathType)
+    {
+        std::string finalPath = "";
+        std::string homePath = "";
+        std::string strPathType = "";
+		struct passwd* pwd = getpwuid(getuid());
```
trzy, line 53: Indentation.

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+            {
+                finalPath = "";
+            }
+            else
+            {
+                // If directory doesn't exist, create it
+                if (!FileSystemPath::PathExists(strPathType)) mkdir(strPathType.c_str(), 0775);
+                finalPath = strPathType;
+            }
+
```
trzy, line 102: Remove empty line.

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+namespace FileSystemPath
+{
+    // Checks if a directory exists (returns true if exists, false if it doesn't)
+    bool PathExists(std::string fileSystemPath)
+    {
+        bool pathExists = false;
+        struct stat pathInfo;
+
+        if (stat(fileSystemPath.c_str(), &pathInfo) == 0 && S_ISDIR(pathInfo.st_mode))
```
trzy, line 38: Not a huge deal since this is intended to run on UNIX systems but does the standard library still lack such functions?

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+        {
+            // Use current directory
+            if (pathType == Screenshots || pathType == Log)
+            {
+                finalPath = "";
+            }
+            else
+            {
+                // If directory doesn't exist, create it
+                if (!FileSystemPath::PathExists(strPathType)) mkdir(strPathType.c_str(), 0775);
```
trzy, line 99: Formatting: Move mkdir to its own line enclosed with braces as elsewhere.

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+        }
+        // On Linux one may want to follow the XDG base directory specs (https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)
+        else
+        {
+            // Use $HOME/.config/supermodel or $HOME/.local/share/supermodel depending on the file type
+            if (pathType == Config)
+            {
+                finalPath = Util::Format() << homePath << "/.config/supermodel";
+                // If directory doesn't exist, create it
+                if (!FileSystemPath::PathExists(finalPath)) mkdir(finalPath.c_str(), 0775);
```
trzy, line 123: Prefer if-statements to be multi-line with braces.

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+                // If directory doesn't exist, create it
+                finalPath = Util::Format() << homePath << "/.config/supermodel/Config";
+                if (!FileSystemPath::PathExists(finalPath)) mkdir(finalPath.c_str(), 0775);
```
trzy, line 126: Prefer if-statements to be multi-line with braces.

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+                
+            }
+            else
+            {
+                finalPath = Util::Format() << homePath << "/.local/share/supermodel";
+                // If directory doesn't exist, create it
+                if (!FileSystemPath::PathExists(finalPath)) mkdir(finalPath.c_str(), 0775);
```
trzy, line 133: Prefer if-statements to be multi-line with braces.

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+                // If directory doesn't exist, create it
+                finalPath = Util::Format() << homePath << "/.local/share/supermodel/" << strPathType;
+                if (!FileSystemPath::PathExists(finalPath)) mkdir(finalPath.c_str(), 0775);
```
trzy, line 136: Prefer if-statements to be multi-line with braces.

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+            }
+            
+        }
+        
+        if (finalPath != "") finalPath = Util::Format() << finalPath << "/";
```
trzy, line 141: Prefer if-statements to be multi-line with braces.

Src/OSD/Unix/FileSystemPath.cpp
```diff
@@ -0,0 +1,146 @@
...
+        switch (pathType)
+        {
+            case Analysis:
+                strPathType = "Analysis";
+                break;
+            case Config:
+                strPathType = "Config";
+                break;
+            case Log:
+                strPathType = "Log";
```
trzy, line 65: I'm not sure about this discrepancy in log and screenshot paths on UNIX vs. Windows. Would it be bad to have them follow the Windows convention?

Alternatively, we could add a Screenshot directory to the base of the Supermodel repo. Logs I'm not sure about. Need to think about whether it warrants its own directory. My gut instinct is 'no'.

trzy (changes requested):
Overall looks good. A few things to address that should be relatively simple. Let me know if you need help with the screenshot filename formatting.

The issue of the log and screenshot directories will need a decision. If we introduce a Screenshots directory, I think the Makefile should create it and add a DIR.txt file describing it, at least on Windows where the directories are not being auto-created. I forget if this happens now with other requires directories like Saves, NVRAM, etc. This could be part of a separate commit.

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: Fixed.

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: Fixed.

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: Fixed.

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: Fixed.

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: As far as I could research it seems it does. Do you want me to do something about it or it's ok to leave it as is?

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: Fixed.

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: Fixed.

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: Fixed.

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: Fixed.

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: Fixed.

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: Fixed.

Src/OSD/FileSystemPath.h
casasfernando: Fixed.

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: IMHO the whole point of this PR is to be able to find all Supermodel related files (and yes, that includes screenshots and the log) in a predictable common path to avoid having files spread on different places depending on which directory the user was sitting when executing Supermodel. No matter where the user was when he called the binary, he will always find all the files in the same predictable place keeping things nice and tidy.
And this should be the same on every platform if you ask me. :smile: 

Just my two cents.

Src/OSD/SDL/Main.cpp
casasfernando: In all honesty I have no clue. :smile:
I can do some research and see what I can come up with.
Just bear in mind that I wanted to touch the existing code as less as possible due to me not being skilled in C++. "If it works, don't touch it!" :smile:

Src/OSD/Unix/FileSystemPath.cpp
casasfernando: What about using a null string as default case? That would translate to current directory in Windows and Linux (if Config is in current path) or ~/.supermodel or ~/.local/share/supermodel depending on which path exists.

What do you think? It's simpler than bringing down Supermodel on error.

casasfernando:
@trzy just pushed a commit based on some of your comments (already marked them as resolved).
Only three outstanding things to make a decision on, so I can change the code accordingly.

Looking forward your thoughts.

Thanks.

trzy:
Thanks, I will take a closer look this weekend!

trzy:
Sorry for the delay I will try to get to it this week later. Been a really busy weekend. I think we are almost there and I might suggest the Util::Format() modifier we need.

Src/OSD/SDL/Main.cpp
trzy: Okay, let's get this over the finish line. Here is the change that will make this look correct: 

```
void Screenshot()
{
    // Make a screenshot
    time_t now = std::time(nullptr);
    tm* ltm = std::localtime(&now);
    std::string file = Util::Format() << filepath
        << "Screenshot_"
        << std::setfill('0') << std::setw(4) << (1900 + ltm->tm_year)
        << '-'
        << std::setw(2) << (1 + ltm->tm_mon)
        << '-'
        << std::setw(2) << ltm->tm_mday
        << "_("
        << std::setw(2) << ltm->tm_hour
        << '-'
        << std::setw(2) << ltm->tm_min
        << '-'
        << std::setw(2) << ltm->tm_sec
        << ").bmp";

    std::cout << "Screenshot created: " << file << std::endl;
    SaveFrameBuffer(file);
}
```
These std format specifiers are stateful and more than a bit annoying. For example, std::setfill() gets set until it is unset for the entire stream but setw() only affects the following element and is then reset each time. What an awful API! Eventually we could just write a Util::Int() or Util::Uint() that does this automatically but this is fine for now.

Src/OSD/Unix/FileSystemPath.cpp
trzy: Since this is under the OSD/Unix/ directory, then yes, this is fine for now.

trzy (commented):
Looking good. Final stretch here.

casasfernando:
@trzy I just pushed another commit to sort out the screenshot function as requested.

Are we good to go or you want to take another look at this?

Thanks!

trzy:
I think we are good but I'd like to give it a test in Ubuntu. I will try to do that tonight but if I don't get to it by tomorrow night I'll trust it's okay and just merge it in.

casasfernando:
Perfect. I tested it in Ubuntu 20.04 already (config, nvram, screenshots and savestates) successfully so you may want to try a different version like 22.04.

trzy:
I didn't have time to test in Linux and I couldn't build successfully on MacOS but there is one last thing: has the MacOS Makefile been modified to continue to compile? I believe it needs one of the FileSystem.cpp implementations. Probably the UNIX one will work but if you can't test it, then use the default one.

casasfernando:
Sorry for the oversight. Indeed I didn't modify the Makefile for macOS.
Just added a commit to sort that out and also built and tested it successfully on MacBook Pro (M1).
I used the default cpp file, same as Windows, since I'm not sure which would be the correct/expected path for the files in macOS and I prefer to keep it unmodified from the current behavior (same as Windows).

PR #47, opened 2022-11-20
Incomplete texture headers uploaded via FIFO are now discarded

gm-matthew:
Ski Champ sends one more 32-bit word via DMA than it should which was causing Supermodel to erroneously interpret the extra word as another texture

### DIFF
--- a/Src/Model3/Real3D.cpp
+++ b/Src/Model3/Real3D.cpp
@@ -656,9 +656,9 @@ void CReal3D::Flush(void)
   DebugLog("Real3D 88000000 written @ PC=%08X\n", ppc_get_pc());
 
   // Upload textures (if any)
-  if (fifoIdx > 0)
+  if (fifoIdx > 2) // If the texture header/data aren't present, discard the texture (prevents garbage textures in Ski Champ)
   {
-    for (uint32_t i = 0; i < fifoIdx; )
+    for (uint32_t i = 0; i < fifoIdx - 2; )
     {
       uint32_t size = 2+textureFIFO[i+0]/2;
       size /= 4;


Src/Model3/Real3D.cpp
```diff
@@ -656,9 +656,9 @@ void CReal3D::Flush(void)
   DebugLog("Real3D 88000000 written @ PC=%08X\n", ppc_get_pc());
 
   // Upload textures (if any)
-  if (fifoIdx > 0)
+  if (fifoIdx > 2) // If the texture header/data aren't present, discard the texture (prevents garbage textures in Ski Champ)
   {
-    for (uint32_t i = 0; i < fifoIdx; )
+    for (uint32_t i = 0; i < fifoIdx - 2; )
```
trzy, line 661: Are these a legitimate bug in Ski Champ that you think the Real3D is capable of ignoring, an emulation bug in Supermodel causing incomplete textures being written, or is there something besides texture size encoded here (are any high bits set)?

gm-matthew: Pretty sure it's a legitimate bug in Ski Champ. For example it will upload a texture that is 0x2AAB0 bytes in size (verified because each texture is spaced that far apart) but it actually sends 0x2AAB4 bytes via DMA; the last four bytes sent via DMA are actually the first four bytes of the next texture, which Supermodel incorrectly interprets as the size of the next texture. Luckily the next "header" is all zeroes so the garbage texture ends up being size 32x32 at position (0,0) which means it probably doesn't corrupt any actual textures.

trzy: That's really interesting. Bugs are of course totally plausible. These custom systems were probably barely documented at the time and I'm sure the programmers occasionally made silly mistakes that just didn't happen to be fatal.

trzy:
Approved and merging.

Src/Model3/Real3D.cpp
dukeeeey: Did you guys ever see if these textures show up normally on the original hw. This definitely isn't the only game that does buggy things.

trzy:
I think he’s saying the texture either never gets written or if it does, it’s some 32x32 junk in the top left of the texture sheet that presumably isn’t identifiably being used anywhere.Sent from my iPhoneOn Nov 20, 2022, at 5:48 AM, dukeeeey ***@***.***> wrote:﻿
@dukeeeey commented on this pull request.



In Src/Model3/Real3D.cpp:
>    {
-    for (uint32_t i = 0; i < fifoIdx; )
+    for (uint32_t i = 0; i < fifoIdx - 2; )

Did you guys ever see if these textures show up normally on the original hw. This definitely isn't the only game that does buggy things.

dukeeeey:
isn't 0,0 the bottom left? But yes that part of the sheet is almost always used :)

gm-matthew:
It turns out the garbage textures actually have no effect; because the header is all zeros the sixteenBit, writeLowerByte and writeUpperByte flags are all clear and nothing is actually written to texture RAM. I have verified that the texture sheets are identical, with and without the "garbage textures".

PR #48, opened 2022-11-22
Wait until IRQ2 is acknowledged before starting IRQ40 sequence

gm-matthew:
Fixes Ski Champ hanging at FBI screen when region set to USA

### DIFF
--- a/Src/Model3/Model3.cpp
+++ b/Src/Model3/Model3.cpp
@@ -2070,6 +2070,8 @@ void CModel3::RunMainBoardFrame(void)
 		IRQ.Assert(0x02);								// start at 33% of the frame
 
 		// keep running cycles until IRQ2 is acknowledged
+        // Ski Champ can hang if we check the MIDI control port too early
+        // and miss MIDI interrupts pending before the next IRQ2
 		while (IRQ.ReadIRQEnable() & 0x2 && IRQ.ReadIRQState() & 0x2 && dispCycles > 1000)
 		{
 			ppc_execute(1000);
--- a/Src/Model3/Model3.cpp
+++ b/Src/Model3/Model3.cpp
@@ -2070,8 +2070,8 @@ void CModel3::RunMainBoardFrame(void)
 		IRQ.Assert(0x02);								// start at 33% of the frame
 
 		// keep running cycles until IRQ2 is acknowledged
-        // Ski Champ can hang if we check the MIDI control port too early
-        // and miss MIDI interrupts pending before the next IRQ2
+		// Ski Champ can hang if we check the MIDI control port too early
+		// and miss MIDI interrupts pending before the next IRQ2
 		while (IRQ.ReadIRQEnable() & 0x2 && IRQ.ReadIRQState() & 0x2 && dispCycles > 1000)
 		{
 			ppc_execute(1000);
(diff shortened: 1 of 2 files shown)


trzy:
I posted a reply on the dev forum to your description of this problem. Not sure I fully understand what's going on but here is what I wrote:

```
The sound board IRQs have always been a mystery. The current code checks midiCtrlPort & 0x20 to begin firing them. I don't understand what you mean when you say Supermodel starts firing them before the port is set to 0x27? Supermodel only seems to fire them when that port has bit 0x20 set.

If I understand correctly, you are saying that the IRQ02 handler in this game takes an unusually long time and while the IRQ02 line has not been deasserted and its handler is still running, Supermodel pushes through a bunch of IRQ40s and when the IRQ02 handler finally finishes and presumably sets some flag, the game then waits for those IRQs (and already missed them)?
```

If my understanding is correct, then isn't this something of a hack? I'm not sure whether our problem is that the display timing is wrong or whether the sound board only fires IRQs during a particular part of the frame even if bit 0x20 of the control port is set. I suppose it's not impossible that the IRQ mechanism is somehow tied to frame timing (particularly VBL) but it does strike me as unlikely. I would think the mechanism is asynchronous and that the whole point of the control port is to tell the sound board when to process commands without needing additional circuitry driven by VBL.

gm-matthew:
In a sense Supermodel's existing approach is also a hack, since it starts the IRQ40 sequence at a fixed time after IRQ2 (2.5% of the frame time). I'm fairly sure on real hardware the soundboard is unaffected by frame timing and would be able to trigger an IRQ40 at any time (as long as it is enabled in the IRQ controller, of course).

Right now for each frame Supermodel performs this sequence: run 33% of the frame, fire IRQ2, run 2.5% of the frame, start the IRQ40 sequence if bit 0x20 of the control port is set, fire IRQ1, IRQ4 and IRQ8 simultaneously, run the rest of the frame. Allowing IRQ40s to be fired independent of frame timing would require reimplementing the RunMainBoardFrame() function so that the control port is checked every x number of cycles; I did consider this approach before deciding on my eventual solution which is a lot easier.

My original fix was pretty much just copy-pasting the IRQ40 sequence code to before IRQ2 as well; now THAT was a hack :-)

trzy:
I replied via email but for some reason GitHub apparently didn't get it and post it here: 

Good points!

I forget exactly how the IRQ assert logic works but on real CPUs, IRQs are usually level or edge-triggered. Let's say they are level-triggered, meaning that as long as that line is held high, IRQ  will keep being generated until explicitly deasserted. The way to do this would be to have a CPU interface that exposes a stateful INT signal. While high, this should constantly generate external interrupt exceptions. This would be checked on each instruction. I think MAME uses a system like this when possible.

It would be interesting to see if such an approach fixes the problem as well. The performance impact should be small (an additional if-statement per instruction, which will almost always evaluate false; branch prediction logic should effectively predict it away). But that's a much bigger effort.

We can go ahead and merge this but I do have one more request: would you be able to write up a description of what's going on, either as a comment block in that function or as a couple paragraphs somewhere in the big comment block at the top of Model3.cpp? At least for me, it's easy to forget what's going on after coming back to this stuff months later.

gm-matthew:
Added the comment: "Ski Champ can hang if we check the MIDI control port too early and miss MIDI interrupts pending before the next IRQ2". Hope it's descriptive enough, I didn't want to create a wall of text

trzy:
Sounds good. I'm going to leave your description from the Supermodel forum here as well for posterity (I'm a bit terrified of that Forum going down at some point; need to archive all the content on there somehow).

```
During the frame just before the game hangs, it takes nearly 194,000 cycles for the IRQ2 handler to clear; that's a good 11% of the frame time even with the PowerPC frequency set to 100MHz. What is happening is that at 2.5% of the frame time after IRQ2 Supermodel checks the control port (while the IRQ2 handler is still running), sees 0x6, and concludes that IRQ40 doesn't need to be fired. Later on in the same frame, the IRQ2 handler writes 0x27 to the control port. The game is expecting IRQ40 to be fired before the next IRQ2, but it never does because Supermodel has already checked the control port and decided not to fire IRQ40. I don't know how other games would react to this, but it is fatal to Ski Champ due to the issue I mentioned in the original post.
```

PR #56, opened 2022-12-26
Crosshair

CapitaineSheridan:
Crosshair refactor

-add a new separate class for crosshair
-crosshair coordinates are calculated by matrix instead of recreating every object at correct coordinates
-add ability to scale crosshair by dpi
-add ability to use bitmap crosshair (located in ./Media/). 32bits bmp format + alpha
-cmd line "-bitmapcrosshair" or "-vectorcrosshair" and/or BitmapCrosshair=0|1 in config file
-these changes are only for lost world game with Crosshairs=1|2|3

don't forget to copy the 2 crosshair images in Supermodel/Media folder

note : supermodel_build_bot.py needs to be edited to include media folder with its 3 files


Src/OSD/FileSystemPath.h
```diff
@@ -32,7 +32,7 @@
 
 namespace FileSystemPath
 {
-    enum PathType { Analysis, Config, Log, NVRAM, Saves, Screenshots }; // Filesystem path types
+    enum PathType { Analysis, Config, Log, NVRAM, Saves, Screenshots, Media }; // Filesystem path types
```
trzy, line 35: I think I prefer "Assets" to "Media", which I see commonly used in e.g. web development for asset files that are used by the program.

Src/OSD/SDL/Crosshair.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+#include "Crosshair.h"
+#include "Supermodel.h"
+#include "Graphics/New3D/New3D.h"
+#include "OSD/FileSystemPath.h"
+#include "SDLIncludes.h"
+#include <GL/glew.h>
+#include <vector>
+
+bool CCrosshair::init()
```
trzy, line 30: init() -> Init()

Src/OSD/SDL/Crosshair.cpp
```diff
@@ -0,0 +1,244 @@
...
+{
+    const std::string P1CrosshairFile = Util::Format() << FileSystemPath::GetPath(FileSystemPath::Media) << "p1crosshair.bmp";
```
trzy, line 32: Naming convention: P1CrosshairFIle -> p1CrosshairFile. Please use pascalCase for variables and CamelCase for functions.

Src/OSD/SDL/Crosshair.cpp
```diff
@@ -0,0 +1,244 @@
...
+    const std::string P2CrosshairFile = Util::Format() << FileSystemPath::GetPath(FileSystemPath::Media) << "p2crosshair.bmp";
+
+	IsBitmapCrosshair = m_config["BitmapCrosshair"].ValueAs<bool>();
+    xRes = m_config["XResolution"].ValueAs<unsigned>();
+    yRes = m_config["YResolution"].ValueAs<unsigned>();
+    a = (float)xRes / (float)yRes;
+
+	SDL_Surface* SurfaceCrosshairP1 = SDL_LoadBMP(P1CrosshairFile.c_str());
```
trzy, line 40: SurfaceCrosshairP1, SurfaceCrosshairP2 -> surfaceCrosshairP1, surfaceCrosshairP2

Src/OSD/SDL/Crosshair.cpp
```diff
@@ -0,0 +1,244 @@
...
+    m_vbo.BufferSubData(0, count * sizeof(BasicVertex), verts.data());
+
+    glBindVertexArray(m_vao);
+    glDrawArrays(GL_TRIANGLES, 0, count);
+    glBindVertexArray(0);
+
+    m_shader.DisableShader();
+}
+
+void CCrosshair::DrawBitmapCrosshair(New3D::Mat4 matrix, float x, float y, int TextureNum)
```
trzy, line 198: Why do we need separate DrawCrosshair and DrawBitmapCrosshair? Let's unify them into a single DrawCrosshair() call that can check whether a bitmap or vector crosshair should be drawn. You can have separate private methods for each that are called. Also, instead of passing r, g, b to DrawCrosshair(), you could probably pass a player index and that should be sufficient for determining color for vectored crosshair and texture num for bitmap crosshair. This will simplify Main.cpp as well.

Src/OSD/SDL/Crosshair.h
```diff
@@ -0,0 +1,80 @@
...
+#define INCLUDED_CROSSHAIR_H
+
+#include "Supermodel.h"
+#include "Graphics/New3D/New3D.h"
+
+class CCrosshair
+{
+private:
+	const Util::Config::Node& m_config;
+	bool IsBitmapCrosshair=false;
```
trzy, line 32: IsBitmapCrosshair -> m_isBitmapCrosshair
Use m_pascalCase for private member variables.

Src/OSD/SDL/Crosshair.h
```diff
@@ -0,0 +1,80 @@
...
+	GLuint CrosshairtxID[2] = { 0 };
```
trzy, line 33: CrosshairtxID -> m_crosshairTexId

Src/OSD/SDL/Crosshair.h
```diff
@@ -0,0 +1,80 @@
...
+	int P1CrosshairW = 0, P1CrosshairH = 0, P2CrosshairW = 0, P2CrosshairH = 0;
```
trzy, line 34: P1CrosshairW -> m_p1CrosshairW, etc.

Src/OSD/SDL/Crosshair.h
```diff
@@ -0,0 +1,80 @@
...
+	float diagdpi = 0.0f, hdpi = 0.0f, vdpi = 0.0f;
```
trzy, line 35: All of the members defined here should be changed to conform to: m_pascalCase.

As a side note, when the variable name includes an acronym, like DPI, the usual convention is to stylize it as: m_dpi or m_vDpi.

Src/OSD/SDL/Crosshair.h
```diff
@@ -0,0 +1,80 @@
...
+		float x, y, z;
+	};
+
+	struct UVCoords
+	{
+		UVCoords(float x, float y) : x(x), y(y) {}
+		float x, y;
+	};
+
+	std::vector<BasicVertex> verts;
```
trzy, line 59: verts -> m_verts, UVCoord -> m_uvCoord

Src/OSD/SDL/Crosshair.h
```diff
@@ -0,0 +1,80 @@
...
+	std::vector<UVCoords> UVCoord;
+
+	GLSLShader m_shader;
+	VBO m_vbo;
+	VBO m_textvbo;
+	GLuint m_vao = 0;
+	int m_TextureCoordsCount = 0;
+	const char* vertexShader;
```
trzy, line 67: vertexShader -> m_vertexShader, fragmentShader ->m_fragmentShader

Src/OSD/SDL/Crosshair.h
```diff
@@ -0,0 +1,80 @@
...
+	const char* fragmentShader;
+
+	const int MaxVerts = 1024;  // per draw call
```
trzy, line 70: This is fine since it is a const. I am undecided as to whether MaxVerts or MAX_VERTS should be used but for now can leave as-is.

Src/OSD/SDL/Main.cpp
```diff
@@ -105,6 +106,12 @@ static unsigned  xOffset, yOffset;      // offset of renderer output within Open
 static unsigned  xRes, yRes;            // renderer output resolution (can be smaller than GL viewport)
 static unsigned  totalXRes, totalYRes;  // total resolution (the whole GL viewport)
 
+/*
+ * Crosshair stuff
+ */
+static bool IsBitmapCrosshair = false;
```
trzy, line 112: For static variables, I strongly prefer s_pascalCase. So:
IsBitmapCrosshair -> s_isBitmapCrosshair. However, since CCrosshair has this information, you don't need it here at all, right? I think this should be eliminated as per my other comments in Crosshair.cpp.

Src/OSD/SDL/Main.cpp
```diff
@@ -105,6 +106,12 @@ static unsigned  xOffset, yOffset;      // offset of renderer output within Open
 static unsigned  xRes, yRes;            // renderer output resolution (can be smaller than GL viewport)
 static unsigned  totalXRes, totalYRes;  // total resolution (the whole GL viewport)
 
+/*
+ * Crosshair stuff
+ */
+static bool IsBitmapCrosshair = false;
+CCrosshair* Crosshair = nullptr;
```
trzy, line 113: change to: static CCrosshair *s_crosshair = nullptr;

Src/OSD/SDL/Main.cpp
```diff
@@ -970,11 +867,30 @@ static void UpdateCrosshairs(uint32_t currentInputs, CInputs *Inputs, unsigned c
     GunToViewCoords(&x[1], &y[1]);
 	offscreenTrigger[1] = (Inputs->trigger[1]->offscreenValue) > 0;
   }
+
   // Draw visible crosshairs
-  if ((crosshairs & 1) && !offscreenTrigger[0])  // Player 1
-    DrawCrosshair(m,x[0], y[0], 1.0f, 0.0f, 0.0f);
-  if ((crosshairs & 2) && !offscreenTrigger[1])  // Player 2
-    DrawCrosshair(m,x[1], y[1], 0.0f, 1.0f, 0.0f);
+  if (!IsBitmapCrosshair)
```
trzy, line 872: See my comments in Crosshair.cpp. I think you can eliminate this variable and test entirely. CCrosshair knows if it is bitmap or not and it is dangerous to have the same state variable in two places (it could get out of sync if someone made a change).

We have hard-coded player 1 to be red and player 2 to be green in the vector case. This can simply be moved into CCrosshair and you can have a single DrawCrosshair() call here for both bitmap and vector crosshairs.

Src/OSD/SDL/Main.cpp
```diff
@@ -1721,6 +1638,8 @@ static void Help(void)
   puts("  -show-fps               Display frame rate in window title bar");
   puts("  -crosshairs=<n>         Crosshairs configuration for gun games:");
   puts("                          0=none [Default], 1=P1 only, 2=P2 only, 3=P1 & P2");
+  puts("  -vectorcrosshair        Use built-in crosshair [Default]");
```
trzy, line 1641: Can we change this to: -crosshair-style=<s>, where style is either "vector" or "bmp" (Default: bmp).

Help text should be something like:
```
puts("  -crosshair-style=<s>    Crosshair style: vector or bmp. [Default: vector]");
```

If you do this, change BitmapCrosshair to CrosshairStyle and set it to a default "vector":

```
config.Set("CrosshairStyle", "vector");
```

trzy (changes requested):
This looks really good. I left a lot of comments but really only three things need to be changed:

1. Variable and function naming convention needs to be standardized. I'm probably going to be more strict about this because I think we should move toward a more uniform coding style over time. I assume that you know how to quickly rename things in Visual Studio: put the cursor on a variable name and press Ctrl-R Ctrl-R. You can then easily rename the variable everywhere in the project.

2. I think IsBitmapCrosshair in Main.cpp is totally redundant and you can eliminate it. This also will let you have only one single public DrawCrosshair() method. CCrosshair already knows what kind of crosshair it is, so it should make all style decisions.

3. Personally, I think a single command line option rather than two would be tidier.

Src/OSD/SDL/Main.cpp
```diff
@@ -925,7 +811,7 @@ static void PrintGLError(GLenum error)
 }
 */
 
-static void UpdateCrosshairs(uint32_t currentInputs, CInputs *Inputs, unsigned crosshairs)
+static void UpdateCrosshairs(uint32_t currentInputs, CInputs *Inputs, unsigned crosshairs, std::string crosshairStyle)
```
trzy, line 814: I didn't notice this earlier but I think this should also be folded into CCrosshair and renamed to Update. Then, you can make even DrawCrosshairs() private and only call s_crosshairs->Update(...). And no more need to pass crosshairStyle parameter because you already have it in CCrosshair.

This means the entire "UI rendering" section of Main.cpp can disappear.

Src/OSD/SDL/Crosshair.cpp
```diff
@@ -0,0 +1,247 @@
...
+#include "SDLIncludes.h"
+#include <GL/glew.h>
+#include <vector>
+
+bool CCrosshair::Init()
+{
+    const std::string p1CrosshairFile = Util::Format() << FileSystemPath::GetPath(FileSystemPath::Assets) << "p1crosshair.bmp";
+    const std::string p2CrosshairFile = Util::Format() << FileSystemPath::GetPath(FileSystemPath::Assets) << "p2crosshair.bmp";
+
+    m_isBitmapCrosshair = m_config["CrosshairStyle"].ValueAs<std::string>();
```
trzy, line 35: I think this should be a bool and you should check the value of CrosshairStyle here. But also don't forget that because this is a string, we need to account for capitalization (e.g., "BMP" vs. "bMp") as well as invalid settings. Util/Format.h has some helpful functions there. I would do it like this:

```
std::string crosshairStyle = Util::ToLower(Util::TrimWhiteSpace(m_config["CrosshairStyle"].ValueAs<std::string>()));
if (crosshairStyle == "bmp")
{
    m_isBitmapCrosshair = true;
}
else if (crosshairStyle == "vector")
{
   m_isBitmapCrosshair = false;
}
else
{
    ErrorLog("Invalid crosshair style '%s'. Must be 'vector' or 'bmp'.", m_config["CrosshairStyle"].ValueAs<std::string>().c_str());
    return FAIL;
}
```

}

trzy (changes requested):
Almost there! Your refactor exposed one more thing that can be refactored to simplify everything.

PR #58, opened 2022-12-29
Add missing filename to ErrorLog call.

AaronBPaden:
Simple fix for a problem I noticed when tracking down a Linux packaging bug.


trzy:
Thanks -- good catch!

PR #62, opened 2023-02-13
Add multiple independent mice support for Linux

DirtBagXon:
This removes SDL_Mouse polling and replaces with
 Ryan C. Gordon's ManyMouse libraries: https://github.com/icculus/manymouse

 This affects Linux only. It uses X11 XInput and evdev polling when available.

 Replicates rawinput (Windows) in Linux.

 2 player gun games now work, as can be seen in: https://youtu.be/qplfGymaDPI

### DIFF
--- a/Makefiles/Makefile.UNIX
+++ b/Makefiles/Makefile.UNIX
@@ -63,7 +63,7 @@ SDL2_LIBS = `sdl2-config --libs`
 #
 
 PLATFORM_CXXFLAGS = $(SDL2_CFLAGS) -O3
-PLATFORM_LDFLAGS = $(SDL2_LIBS) -lGL -lGLU -lz -lm -lstdc++ -lpthread -lSDL2_net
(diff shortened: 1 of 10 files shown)


DirtBagXon:
Just an update, the issue in Apple was down to _“Input Monitoring”_ permissions, as detailed here:

https://nachtimwald.com/2020/11/08/macos-iohidmanager-permission-issue/

So Apple support for ManyMouse could be easily added on that proviso,
depending on how the game is launched... Or enabled by an new argument....

trzy:
One big concern stands out:
- Can we add this in addition to existing SDL mouse support and make it compile only on Linux systems? My understanding is that Supermodel still works on BSD systems and if this is Linux-specific, that will break that compatibility. Ideally, this would be enabled as an option only under Linux, the way RawInput can be used with Windows only.

DirtBagXon:
Hi, thanks for the feedback, I don't really have time to refactor the Input code at this point, so I'll close this down but if someone else wants to pick up on it, I have no issues. My next idea was just to have it as a `-DSUPERMODEL_MANYMOUSE` compile time flag, but don't think this would satisfy your wants.

Thanks again.

PR #64, opened 2023-03-02
Amelioration for networked multiplayer on single machine

j-lag:
- Allow joystick to be fetch if windows has no focus in SDL input mode
- Add video settings:
      `-window-pos=<x>,<y>`            Position [Default: centered]
      `-borderless`             Windowed mode with no border
These 2 settings are usefull when setting up a multiplayer game on one machine.

Example for 4 windows in fullHD:
`start /D"Master" Supermodel.exe -input-system=sdl dayto2pe.zip -res=960,540 -borderless -window-pos=0,0`
`start /D"Slave1" Supermodel.exe -input-system=sdl dayto2pe.zip -res=960,540 -borderless -window-pos=960,0`
`start /D"Slave2" Supermodel.exe -input-system=sdl dayto2pe.zip -res=960,540 -borderless -window-pos=0,540`
`start /D"Slave3" Supermodel.exe -input-system=sdl dayto2pe.zip -res=960,540 -borderless -window-pos=960,540`

### DIFF
--- a/Src/OSD/SDL/Main.cpp
+++ b/Src/OSD/SDL/Main.cpp
@@ -177,7 +177,7 @@ static void GLAPIENTRY DebugCallback(GLenum source, GLenum type, GLuint id, GLen
 }
 
 // In windows with an nvidia card (sorry not tested anything else) you can customise the resolution.
-// This also allows you to set a totally custom refresh rate. Apparently you can drive most monitors at 
+// This also allows you to set a totally custom refresh rate. Apparently you can drive most monitors at
 // 57.5fps with no issues. Anyway this code will automatically pick up your custom refresh rate, and set it if it exists
 // It it doesn't exist, then it'll probably just default to 60 or whatever your refresh rate is.
 static void SetFullScreenRefreshRate()
@@ -828,7 +828,7 @@ private:
         const char* vertexShader = R"glsl(
 
             #version 410 core
-             
+
             uniform mat4 mvp;
             layout(location = 0) in vec3 inVertices;
 
@@ -841,7 +841,7 @@ private:
         const char* fragmentShader = R"glsl(
 
             #version 410 core
-             
+
             uniform vec4 colour;
             out vec4 fragColour;
 
@@ -1087,15 +1087,10 @@ int Supermodel(const Game &game, ROMSet *rom_set, IEmulator *Model3, CInputs *In
   SDL_SetWindowTitle(s_window, baseTitleStr);
   SDL_SetWindowSize(s_window, totalXRes, totalYRes);
 
-  if (!s_runtime_config["WindowXPosition"].Empty() && !s_runtime_config["WindowYPosition"].Empty())
-  {
-    SDL_SetWindowPosition(s_window, s_runtime_config["WindowXPosition"].ValueAs<unsigned>(), s_runtime_config["WindowYPosition"].ValueAs<unsigned>());
-  }
-  else
-  {
-    SDL_SetWindowPosition(s_window, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);
-  }
-  
+  int xpos = s_runtime_config["WindowXPosition"].Exists() ? s_runtime_config["WindowXPosition"].ValueAs<int>() : SDL_WINDOWPOS_CENTERED;
+  int ypos = s_runtime_config["WindowYPosition"].Exists() ? s_runtime_config["WindowYPosition"].ValueAs<int>() : SDL_WINDOWPOS_CENTERED;
+  SDL_SetWindowPosition(s_window, xpos, ypos);
+
   if (s_runtime_config["BorderlessWindow"].ValueAs<bool>())
   {
     SDL_SetWindowBordered(s_window, SDL_FALSE);
@@ -1642,8 +1637,8 @@ static Util::Config::Node DefaultConfig()
   config.Set("QuadRendering", false);
   config.Set("XResolution", "496");
   config.Set("YResolution", "384");
-  config.Set("WindowXPosition", "NA");
-  config.Set("WindowYPosition", "NA");
+  config.SetEmpty("WindowXPosition");
+  config.SetEmpty("WindowYPosition");
   config.Set("FullScreen", false);
   config.Set("BorderlessWindow", false);
 
@@ -1725,7 +1720,7 @@ static void Help(void)
   puts("");
   puts("Video Options:");
   puts("  -res=<x>,<y>            Resolution [Default: 496,384]");
-  puts("  -window-pos=<x>,<y>     Window position [Default: centered]");  
+  puts("  -window-pos=<x>,<y>     Window position [Default: centered]");
   puts("  -window                 Windowed mode [Default]");
   puts("  -borderless             Windowed mode with no border");
   puts("  -fullscreen             Full screen mode");
@@ -1966,13 +1961,11 @@ static ParsedCommandLine ParseCommandLine(int argc, char **argv)
           }
           else
           {
-              unsigned  x, y;
-              if (2 == sscanf(&argv[i][4], "=%u,%u", &x, &y))
+              int xpos, ypos;
+              if (2 == sscanf(&argv[i][11], "=%d,%d", &xpos, &ypos))
               {
-                  std::string xres = Util::Format() << x;
-                  std::string yres = Util::Format() << y;
-                  cmd_line.config.Set("WindowXPosition", xres);
-                  cmd_line.config.Set("WindowYPosition", yres);
+                  cmd_line.config.Set("WindowXPosition", xpos);
+                  cmd_line.config.Set("WindowYPosition", ypos);
               }
               else
               {
--- a/Src/OSD/SDL/Main.cpp
+++ b/Src/OSD/SDL/Main.cpp
@@ -1087,11 +1087,10 @@ int Supermodel(const Game &game, ROMSet *rom_set, IEmulator *Model3, CInputs *In
   SDL_SetWindowTitle(s_window, baseTitleStr);
   SDL_SetWindowSize(s_window, totalXRes, totalYRes);
 
-  if (  s_runtime_config["Xpos"].ValueAs<std::string>() != "NA" &&
-        s_runtime_config["Ypos"].ValueAs<std::string>() != "NA" )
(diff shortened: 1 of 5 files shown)


Src/OSD/SDL/Main.cpp
```diff
@@ -1086,8 +1086,15 @@ int Supermodel(const Game &game, ROMSet *rom_set, IEmulator *Model3, CInputs *In
   sprintf(baseTitleStr, "Supermodel - %s", game.title.c_str());
   SDL_SetWindowTitle(s_window, baseTitleStr);
   SDL_SetWindowSize(s_window, totalXRes, totalYRes);
-  SDL_SetWindowPosition(s_window, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);
+
+  if ( !s_runtime_config["Xpos"].Empty() && !s_runtime_config["Xpos"].Empty())
```
trzy, line 1090: I think you mean to check Xpos and Ypos. I have a few requests here:

1. Remove the leading whitespace in the leading parenthesis.
2. Add braces to the if and else statements. I know this isn't how the rest of the code base looks (yet) but I'd like us to move to that style as it is considered best practice in all C-like language these days to avoid unbraced if-statements and single-line loops.
3. Please change Xpos and Ypos to "WindowXPosition" and "WindowYPosition" to match the verbose format of most other config options.

Src/OSD/SDL/Main.cpp
```diff
@@ -1086,8 +1086,15 @@ int Supermodel(const Game &game, ROMSet *rom_set, IEmulator *Model3, CInputs *In
...
   SDL_SetWindowTitle(s_window, baseTitleStr);
   SDL_SetWindowSize(s_window, totalXRes, totalYRes);
-  SDL_SetWindowPosition(s_window, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);
+
+  if ( !s_runtime_config["Xpos"].Empty() && !s_runtime_config["Xpos"].Empty())
+    SDL_SetWindowPosition(s_window, s_runtime_config["Xpos"].ValueAs<unsigned>(), s_runtime_config["Ypos"].ValueAs<unsigned>());
+  else  
+    SDL_SetWindowPosition(s_window, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);
   
+  if (s_runtime_config["BorderLess"].ValueAs<bool>())
```
trzy, line 1095: 1. Braces around the statement.
2. BorderLess -> BorderlessWindow

Src/OSD/SDL/Main.cpp
```diff
@@ -1708,7 +1719,9 @@ static void Help(void)
   puts("");
   puts("Video Options:");
   puts("  -res=<x>,<y>            Resolution [Default: 496,384]");
+  puts("  -pos=<x>,<y>            Position [Default: centered]");
```
trzy, line 1722: Change -pos to -window-pos please.

Src/OSD/SDL/Main.cpp
```diff
@@ -1936,6 +1950,31 @@ static ParsedCommandLine ParseCommandLine(int argc, char **argv)
           }
         }
       }
+      else if (arg == "-pos" || arg.find("-pos=") == 0)
```
trzy, line 1953: When changing -pos -> -window-pos don't forget to update all occurrences here.

Src/OSD/SDL/Main.cpp
```diff
@@ -1936,6 +1950,31 @@ static ParsedCommandLine ParseCommandLine(int argc, char **argv)
           }
         }
       }
+      else if (arg == "-pos" || arg.find("-pos=") == 0)
+      {
+          std::vector<std::string> parts = Util::Format(arg).Split('=');
+          if (parts.size() != 2)
+          {
+              ErrorLog("'-pos' requires both a X and Y (e.g., '-pos=10,0').");
```
trzy, line 1958: Nitpick: "both a X and Y" -> "both an X and Y position"

Src/OSD/SDL/Main.cpp
```diff
@@ -1936,6 +1950,31 @@ static ParsedCommandLine ParseCommandLine(int argc, char **argv)
...
+              if (2 == sscanf(&argv[i][4], "=%u,%u", &x, &y))
+              {
+                  std::string xres = Util::Format() << x;
+                  std::string yres = Util::Format() << y;
+                  cmd_line.config.Set("Xpos", xres);
+                  cmd_line.config.Set("Ypos", yres);
+              }
+              else
+              {
+                  ErrorLog("'-pos' requires both a X and Y (e.g., '-pos=10,0').");
```
trzy, line 1973: Same here.

trzy (changes requested):

j-lag:
No worries!
Changes done in comit 722c70bf865892fa64cc369e368e2fe170b1b403 !
best regards.

Src/OSD/SDL/Main.cpp
```diff
@@ -1708,7 +1725,9 @@ static void Help(void)
   puts("");
   puts("Video Options:");
   puts("  -res=<x>,<y>            Resolution [Default: 496,384]");
+  puts("  -window-pos=<x>,<y>     Position [Default: centered]");
```
trzy, line 1728: Change description to: Window position [Default: centered]

Src/OSD/SDL/Main.cpp
```diff
@@ -1086,8 +1086,21 @@ int Supermodel(const Game &game, ROMSet *rom_set, IEmulator *Model3, CInputs *In
...
   SDL_SetWindowSize(s_window, totalXRes, totalYRes);
-  SDL_SetWindowPosition(s_window, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);
+
+  if (!s_runtime_config["WindowXPosition"].Empty() && !s_runtime_config["WindowYPosition"].Empty())
+  {
+    SDL_SetWindowPosition(s_window, s_runtime_config["WindowXPosition"].ValueAs<unsigned>(), s_runtime_config["WindowYPosition"].ValueAs<unsigned>());
+  }
+  else
+  {
+      SDL_SetWindowPosition(s_window, SDL_WINDOWPOS_CENTERED, SDL_WINDOWPOS_CENTERED);
```
trzy, line 1096: Looks like an indentation issue on either this line and/or the if-statement above.

trzy (changes requested):
Looks great and thank you for the changes! Just two very small changes (an indentation issue and adding a word the help description) and we should be good to merge.

j-lag:
Done!
Yea my visual studio and I ar used to 4 spaces indents when your code use 2
spaces indents, thats why :)

trzy:
Quick question: you set the default values to NA but that means the line that checks for whether those are empty will always return false. I didn’t anticipate this situation where there is a purely optional config value. Let me test this in a few hours and get back to you. The config system doesn’t necessary need to have defaults set. 

Supermodel will try to cast “NA” to unsigned which will probably be 0. We don’t have a convention of using different types for the same config key. There are really three ways to handle this: 1) a special magic number is used (like SDL has for the window centered positions), which isn’t great, or I think 2) we can avoid setting these values in the default initialization. The values could 3) also be made signed ints and negative values can be interpreted to center the window.

Which do you prefer? I am leaning toward option 2, with options 3 and 1 as my next choices.

If we go with option 2 there should be a comment added near the empty check that default values for WindowXPos and WindowYPos do not exist so empty must be checked for explicitly.

trzy:
Actually I thought of an even better way: the config system should support “empty” values. Empty() means the config key doesn’t exist. We should have a method that checks if it is Empty() or if the value is an Empty() string (0 length or containing only white space). EmptyOrBlank().

I’ll paste some code in an hour or so to show exactly what I mean.

I can then do a follow on PR that changes Empty() too !Exists() and EmptyOrBlank() to Empty(). I think that is probably a more clear design?

j-lag:
Hi!
Actually I first wrote something along the line of if(strcmp(posx , "NA")
!=0 ...)
But when I tested it it crash when not set.
I have not look further into your default param system but I found that the
parameter is empty when not set by the cmdline. Bottom line : it work as
is...

j-lag:
Ok fine, I will not have a lot of time to test in the upcoming days but I
will happily integrate any change you want.

trzy:
Ok, I have made some fixes. Please see my branch (bart/windowpos) and cherry pick the two commits I made into yours (or rebase yours on mine):

```
* bb8b31e  (HEAD -> bart/windowpos, origin/bart/windowpos) (2023-03-04 11:59:49) Bart Trzynadlowski Fixed window position config storage and command line parsing
* 43833a2  (2023-03-04 11:58:52) Bart Trzynadlowski Config nodes: added the ability to clear out node values or create empty leaf nodes
```
Three files were changed: Main.cpp, NewConfig.cpp, NewConfig.h. Here's what I did:

- Made the values ints instead of unsigned because that's what SDL takes.
- Check each of WindowXPosition, WindowYPosition individually and if it does not exist, use SDL_WINDOWPOS_CENTERED for only that axis. This means if you specify one of these options in the config file but not the other, it will still work, but only one axis will be set and the other will be centered.
- Updated NewConfig.cpp/NewConfig.h to support creating empty nodes. The key will be present in the config tree but there will be no value, so Empty() returns true and Exists() returns false. No change to the API as a result of this!

This is different than what I proposed. I had to refresh myself on my complicated config system :) But it's a better solution. Basically, the config system works like this:

- It's a nested tree of keys and values.
- Values can be almost any type. They are *not* strings underneath (I had forgotten about this). In fact, they are whatever type is used to initially set them.
- When loading from INI or XML, the values will indeed be loaded as strings. But otherwise, they can be set directly to ints or floats or whatever, and that will be the underlying representation. ValueAs<T>() only performs a conversion if it is not already stored as T.
- ValueAs<T>() will throw an exception if there is no value in the node.
- ValueAsDefault<T>(T default) will not, instead it will behave like ValueAs<T>() if a value exists, otherwise it will return the default you specify. It never throws an exception.
- I've added a SetEmpty() method that can be used to create empty keys or clear existing ones and I use this for WindowXPosition, etc.

Hopefully this makes more sense.

j-lag:
Cool I will check that tomorrow after work!

trzy:
Thanks a lot!

j-lag:
👍

PR #65, opened 2023-03-11
Sinden lightgun

j-lag:
Draw a white border around image to be compatible with "Sinden-like" video based lightgun.
```
-draw-white-border
```


trzy:
I'm working on helping the crosshairs PR get integrated which removes UI rendering from Main.cpp entirely. I wonder if this could be integrated into the CCrosshairs class that will be created (should be pretty easy). Also wonder whether once that change is in, CCrosshairs could be renamed to CLightgunUI and this could be inserted in the update method there.

trzy:
Let's wait a couple days until that change is merged and then assess the best course of action. Should be a small modification either way.

j-lag:
No worries man...

trzy:
Now that the crosshairs change has been merged in, I think this belongs inside of CCrosshair. We may also want to rename CCrosshair to CLightgunUI.

j-lag:
Ok will look at the lastest stuff when I got time/
I just got a ps5 so It may take a bit :)

PR #66, opened 2023-03-14
Autobuild script update following crosshair update

CapitaineSheridan:


trzy:
Looks good. Unfortunately I need to manually update the PC this is running on tomorrow so the next build will be missing these. I’ll probably manually fix the download page.

PR #67, opened 2023-03-15
Update README.txt

ToBul:
Fix old typos, leaving alone the American English.
Remove the "Virtua Striker 2 '99" unlock code. I don't think it works, others have also tried,
https://www.supermodel3.com/Forum/viewtopic.php?f=3&t=2096&p=19784#p19784


trzy:
Wow. Table of 'Conents' -- how did that go unnoticed for 12 years? 🗡️ 
Thanks!

PR #75, opened 2023-05-15
GitHub workflows

h0tw1r3:
I was using this on my personal fork for awhile now (syncing with svn).

Creates artifacts for each of the supported platforms.

It doesn't contain a workflow for releases, but that should be really easy.


h0tw1r3:
Here's what the workflows looks like in action:
https://github.com/h0tw1r3/model3emu/actions/runs/5060971437

trzy:
Ok. I need to get on this. Maybe I'll try tomorrow. How does this differ from the other PR from December that is still outstanding? If I understand correctly, you have Linux, Windows, *and* MacOS, and it looks like you are generating artifacts, too. Is the best way to test to fork the repo privately and edit there?

h0tw1r3:
@trzy most convenient way to test is to add a commit to this branch, and open a pull request to your personal fork. You should then see a **PR tests** workflow under Actions.

h0tw1r3:
Also adding a workflow dispatch trigger. Merge into the primary branch of your personal repo to test the ability to trigger the workflow ad-hoc.

trzy:
Ok, I really need to get on this and test it out. Can you give me an overview here? If it produces build artifacts that are identical to what I produce with my build bot (including properly packaging the ZIP file with all the required subdirectories, README.txt, LICENSE.txt, etc.), then we could really use this to distribute binaries.

h0tw1r3:
@trzy Yes, that was my intention. The builds contain everything you put in your releases.

WingofaGriffin:
@h0tw1r3 I also tried making actions to automate it ( as seen in the draft PR ), but ran into errors with it not being able to find SDL2.h as described here: https://github.com/trzy/Supermodel/issues/135

How were you able to get around these errors, and do you have a link to your fork so I can test the binaries out?

h0tw1r3:
> @h0tw1r3 I also tried making actions to automate it ( as seen in the draft PR ), but ran into errors with it not being able to find SDL2.h as described here: #135
> 
> How were you able to get around these errors, and do you have a link to your fork so I can test the binaries out?

Opened this almost 10 months ago, I don't recall how I got around any errors, or that there were errors.

https://github.com/h0tw1r3/model3emu/actions/

WingofaGriffin:
> Opened this almost 10 months ago, I don't recall how I got around any errors, or that there were errors.
> 
> https://github.com/h0tw1r3/model3emu/actions/

Thanks. Taking a look, it seems like the solution for MacOS was to manually download and direct the makefile to SDL2 files, which feels a bit hacky to me, as the version of SDL2 is hard coded into the action. I'm curious if we can direct it to the homebrew installation we can perform with actions to something like `/opt/homebrew/include/SDL2` which I've tried but failed at.

Additionally, upon downloading the artifact, it looks like you're zipping the tarball somehow. This seems redundant.

I'm not running a Windows machine right now to validate that build, but this does look like a better more robust implementation than mine otherwise. Good work.

.github/workflows/build.yml
```diff
@@ -0,0 +1,103 @@
...
+          curl -sLo SDL2_net.dmg https://github.com/libsdl-org/SDL_net/releases/download/release-2.2.0/SDL2_net-2.2.0.dmg
+          hdiutil attach -quiet SDL2.dmg
+          hdiutil attach -quiet SDL2_net.dmg
+          mkdir Frameworks
+          cp -r /Volumes/SDL2/SDL2.framework Frameworks/
+          cp -r /Volumes/SDL2_net/SDL2_net.framework Frameworks/
+          xattr -dr com.apple.quarantine Frameworks/SDL2.framework
+          xattr -dr com.apple.quarantine Frameworks/SDL2_net.framework
+          hdiutil detach -quiet /Volumes/SDL2
+          hdiutil detach -quiet /Volumes/SDL2_net
```
WingofaGriffin, line 69: This feels messy and hard coded. Ideally we'd use homebrew to install SDL2, but if we have to do this can we dynamically get the SDL2 version instead?

.github/workflows/build.yml
```diff
@@ -0,0 +1,103 @@
...
+        if: startsWith(matrix.os, 'windows')
+        uses: msys2/setup-msys2@v2
+        with:
+          update: true
+          install: >-
+            make
+            zip
+            mingw64/mingw-w64-x86_64-gcc
+            mingw64/mingw-w64-x86_64-SDL2
+            mingw64/mingw-w64-x86_64-SDL2_net
```
WingofaGriffin, line 80: Does this still work despite just using `make` instead of `pacman -S mingw64/mingw-w64-x86_64-make`?

.github/workflows/build.yml
```diff
@@ -0,0 +1,103 @@
...
+          cp Docs/LICENSE.txt Docs/README.txt .
+          echo "NMRAM files go here." > NVRAM/DIR.txt
+          echo "Save states go here." > Saves/DIR.txt
+          echo "Recommended (but not mandatory) location for ROM sets." > ROMs/DIR.txt
+          FILES="supermodel${{ matrix.outfile_ext }} LICENSE.txt README.txt Config/ Assets/ NVRAM/ Saves/ ROMs/ ${{ matrix.archive_files }}"
+          if [[ "${{ matrix.archive }}" == "tar" ]] ; then
+            tar -cvf supermodel-${{ github.sha }}.tar $FILES
+          elif [[ "${{ matrix.archive }}" == "zip" ]] ; then
+            zip -r supermodel-${{ github.sha }}.zip $FILES
+          fi
```
WingofaGriffin, line 99: This is zipping the tarball

Makefiles/Makefile.OSX
```diff
@@ -65,8 +65,8 @@ endif
 #	OSX-specific
 #
 
-PLATFORM_CXXFLAGS = $(SDL_CFLAGS) -DSUPERMODEL_OSX -F/Library/Frameworks/ -O3
-PLATFORM_LDFLAGS = $(SDL_LIBS) -lz -lm -lstdc++ -F/Library/Frameworks/
+PLATFORM_CXXFLAGS = -F./Frameworks $(SDL_CFLAGS) -DSUPERMODEL_OSX
+PLATFORM_LDFLAGS = -F./Frameworks $(SDL_LIBS) -lz -lm -lstdc++ -Wl,-rpath,@executable_path/../Frameworks -Wl,-rpath,@executable_path/Frameworks
```
WingofaGriffin, line 69: See the yml file for comments on location of the Frameworks dir

.github/workflows/build.yml
h0tw1r3: Disagree.
This is a standard workflow for a developer who would build the software. Aside from install path, it follows the README build instructions.
Why is homebrew preferred? It's a completely unnecessary dependency.

Makefiles/Makefile.OSX
h0tw1r3: Again, I completely disagree with explicitly creating a homebrew requirement for built binaries, and it deviates from the build documentation.
 
This corrects the need for an end-user to download the SDL framework, as documented in the last compiled download on the supermodel3.com website.

_On Mac, applications are expected to ship with the dependencies they need to run._

.github/workflows/build.yml
h0tw1r3: This is a known artifact bug. Nothing to do here.
https://github.com/actions/upload-artifact/issues/109

Keep in mind this is also for a CI build pipeline, **not** a release pipeline.

In a follow-up PR, this workflow will be converted to a reusable workflow, which takes the final artifact and creates a tagged release, which does not have the limitations of the actions artifacts API.

.github/workflows/build.yml
h0tw1r3: Of course, there is a validate step which runs the binary:
https://github.com/h0tw1r3/model3emu/actions/runs/8116697347/job/22187293159#step:7:1

.github/workflows/build.yml
WingofaGriffin: Understood. I think I was just more put aback by the hard coded version number. Given that SDL2 doesn't have major releases that often though I assume it's fine.

Homebrew is included by default on the Actions VM and IMO has become the defacto way to install dev dependencies in MacOS for compilation. As I am not the project dev, @trzy should decide how they want to move forward with this

h0tw1r3: @WingofaGriffin my personal opinion is that Homebrew is a crutch, and should **never be requirement** for projects like this. There's no reason not to support building with Homebrew, but shipping binaries linked to Homebrew assets is a support minefield.

h0tw1r3: > Understood. I think I was just more put aback by the hard coded version number. Given that SDL2 doesn't have major releases that often though I assume it's fine.
> 

That's fair. I'll add support for building against the latest SDL2 release.

Makefiles/Makefile.OSX
WingofaGriffin: In this same vein, I'm curious if it makes sense to bundle the linux one as an AppImage? Probably can be something separate, but thinking out loud.

h0tw1r3: > In this same vein, I'm curious if it makes sense to bundle the linux one as an AppImage? Probably can be something separate, but thinking out loud.

Unrelated to this PR. I suggest opening an issue.

trzy:
Sorry, haven't been following along. What is the current state of this? I know there are some macOS Makefile issues apparently wrt SDL2. I'd like to try to find time to tackle this in the next week and get it merged in.

Makefiles/Makefile.Win32
```diff
@@ -82,8 +82,8 @@ endif
 #
 # SDL2
 #
-SDL2_LIBS =  -lmingw32 -lSDL2main -lSDL2 -Wl,--no-undefined -lshell32 -lsetupapi -ladvapi32 -luuid -lversion -limm32 -lwinmm -lgdi32 -luser32 -lm -pipe
-SDL2_CFLAGS =
+SDL2_LIBS = `sdl2-config --static-libs`
+SDL2_CFLAGS = `sdl2-config --cflags | sed 's/-Dmain=SDL_main//'`
```
trzy, line 86: Why is removing SDL_main required?

Makefiles/Makefile.OSX
trzy, line 69: These don't work on my Mac machine at all. Here is what I need to make it work on my machine:

```
PLATFORM_CXXFLAGS = $(SDL_CFLAGS) -DSUPERMODEL_OSX -F/Users/bart/Library/Frameworks/ -O3
PLATFORM_LDFLAGS = $(SDL_LIBS) -lz -lm -lstdc++ -F/Users/bart/Library/Frameworks/ -Wl,-rpath,/Users/bart/Library/Frameworks
```

I installed SDL2 with brew.

trzy: Actually, here is something that works. Can you test on your machine? Note that no rpath is needed here:

```
#
# SDL
#

SDL2_CFLAGS = `sdl2-config --cflags`
SDL2_LIBS = `sdl2-config --libs`

#
#	OSX-specific
#

PLATFORM_CXXFLAGS = $(SDL2_CFLAGS) -O3
PLATFORM_LDFLAGS = $(SDL2_LIBS) -lz -lm -lstdc++ -framework AGL -framework OpenGL -framework GLUT -framework Cocoa
```

trzy: I get a crash in CCrosshair::Init() no matter what when using the legacy engine. Need to debug that. Lots of shader errors too.

Makefiles/Makefile.Win32
h0tw1r3: because it conflicts with https://github.com/trzy/Supermodel/blob/dec85032ba4f0b37996da26157b6aa0edfc9196c/Src/OSD/SDL/Main.cpp#L1901
only applies to windows.

Makefiles/Makefile.OSX
h0tw1r3: > These don't work on my Mac machine at all. Here is what I need to make it work on my machine:

> I installed SDL2 with brew.

Generally I would recommend *not* directly supporting homebrew. The SDL Framework from the vendor is easy to compile with, and eliminates the myriad of homebrew dependencies when trying reproduce an issue.
It's also a win for build automation, pull in the framework, compile, release. Homebrew would require compilng SDL before building supermodel 😞

`/Library/Frameworks` should be automatically in the frameworks path, which is why I removed it. So the documented way of building in the `README.md` should work. If you're saying it doesn't, I'll take a look.

I recommend creating a `Frameworks` directory in the source directory and copying the SDL frameworks into that (which is exactly how CI builds it). This follows how Apple intends 3rd parties to build and ship applications.

h0tw1r3: > I get a crash in CCrosshair::Init() no matter what when using the legacy engine. Need to debug that. Lots of shader errors too.
> 
> EDIT: The issue is that this function is missing (it resolves to address 0x0 and thus a segfault):
> 
> ```
> glGenVertexArrays(1, &m_vao);
> ```
> 
> I've read online that <OpenGL/gl3.h> should be included but this does not fix it. Any ideas?

What's version of OSX? (I'm building on 14 Sonoma)

h0tw1r3: > Actually, here is something that works. Can you test on your machine? Note that no rpath is needed here:
> 
> ```
> #
> # SDL
> #
> 
> SDL2_CFLAGS = `sdl2-config --cflags`
> SDL2_LIBS = `sdl2-config --libs`
> 
> #
> #	OSX-specific
> #
> 
> PLATFORM_CXXFLAGS = $(SDL2_CFLAGS) -O3
> PLATFORM_LDFLAGS = $(SDL2_LIBS) -lz -lm -lstdc++ -framework AGL -framework OpenGL -framework GLUT -framework Cocoa
> ```

Doesn't work for me as I don't use Homebrew, and the Framework only supports cmake.

trzy: > Shaders aren't going to work as version 4.10 is required all over the place (outside of New3d).
> And isn't a GL context version of >= 3 required for glGenVertexArrays and friends?


If the legacy engine was broken on macOS, that's fine. We should remove it altogether as an option under macOS with an #ifdef I suppose.

trzy: > > Actually, here is something that works. Can you test on your machine? Note that no rpath is needed here:
> > ```
> > #
> > # SDL
> > #
> > 
> > SDL2_CFLAGS = `sdl2-config --cflags`
> > SDL2_LIBS = `sdl2-config --libs`
> > 
> > #
> > #	OSX-specific
> > #
> > 
> > PLATFORM_CXXFLAGS = $(SDL2_CFLAGS) -O3
> > PLATFORM_LDFLAGS = $(SDL2_LIBS) -lz -lm -lstdc++ -framework AGL -framework OpenGL -framework GLUT -framework Cocoa
> > ```
> 
> Doesn't work for me as I don't use Homebrew, and the Framework only supports cmake.

Okay, I'm going to be in Reno for the Easter holidays and my birthday end of next week so I hope to FINALLY have time to sit down without constant interruptions from other projects to focus on Supermodel for a couple days :)

I'm confused here still. So I read above and you made a compelling argument for not supporting homebrew. That's fine. I hate homebrew personally. But I'm not clear on what here is the problem?

Does `sdl2-config` run? That script should be there no matter what although it may require updating your PATH manually if SDL2 is not installed in the right place. Usually I think sdl2-config is placed somewhere that is already in the path. 

The whole point of sdl2-config is that it should return the necessary information to use in a Makefile. It works on Linux and it technically would work on Windows, too, so long as it is in the path. What exactly are those commands returning on your machine?

I'm almost wondering if we need to have the Makefile fetch SDL2 and build it itself as a subcommand. Maybe that's the best approach. This would be a no-go on Windows probably. It *could* work on Linux. But maybe macOS is a sufficiently controlled environment that this is possible?

h0tw1r3: > Does `sdl2-config` run? That script should be there no matter what although it may require updating your PATH manually if SDL2 is not installed in the right place. Usually I think sdl2-config is placed somewhere that is already in the path.

Nope, sdl2-config is not included with the SDL2 OSX Framework.

> The whole point of sdl2-config is that it should return the necessary information to use in a Makefile. It works on Linux and it technically would work on Windows, too, so long as it is in the path. What exactly are those commands returning on your machine?

All of that is of course true, _except on OSX when using frameworks_ because sdl2-config is unnecessary. You simply add the clang `-F` argument specifying paths to where the frameworks are. It magically figures out the rest without the need for explicitly providing paths to headers or libs.

> I'm almost wondering if we need to have the Makefile fetch SDL2 and build it itself as a subcommand. Maybe that's the best approach. This would be a no-go on Windows probably. It _could_ work on Linux. But maybe macOS is a sufficiently controlled environment that this is possible?

What about a target that downloads the SDL2 framework(s) if they're not in `./Frameworks`?

h0tw1r3: Roughed out support in 1058f3f to automatically download the SDL2 frameworks for building on OSX.

trzy: Will test this weekend! I think downloading the frameworks could be a great solution here if it works reasonably reliably!

trzy:
Pulling SDL2 frameworks seems to be working great! I made some fixes. Can you take a look at my branch bart/macos-makefile-fixes please and pull in the changes? Let me know if anything in there is disagreeable. Here is what I did:

1. I removed the Frameworks directory from the source tree and merged the contents of its Makefile into Makefiles/Makefile.OSX. I feel those rules belong there and that because Frameworks/ is really an artifact of the macOS build process, there's no point in adding it to the shared source tree.
2. I discovered a long-standing flaw in the build system, particularly when using parallel builds (although I don't know if it ever really mattered). Directories BIN_DIR, OBJ_DIR are specified as dependencies on the same level as OBJ_FILES, which is completely wrong. In fact, all object files depend on OBJ_DIR. So I added a phony "directories" target and made every single rule that builds an object file depend on this. This guarantees that OBJ_DIR will be created first, including in parallel builds.
3. I added a PLATFORM_DIRS variable, which allows platform-specific Makefiles to include directory-level targets. This is only needed for macOS and is what now triggers Frameworks/* to be downloaded. This is also parallel-safe: these will be pulled first and *then* all the objects built in parallel.

Once these changes are in, I will do a final test and I think we can integrate the workflow! Then, the Supermodel3.com download page can just pull builds from here as they appear.

trzy:
Ah crap, it's still broken: now it *always* builds everything regardless of whether stale or not. No idea why. This was happening with yours, too.

h0tw1r3:
@trzy Cherry-picked your changes onto the PR.
Cleaned up some of the Framework targets and updated the action workflows to support downloading frameworks with make.
Also added support for downloading the latest version of the frameworks (as the action workflows do) when the source is clean.
Added a release workflow that doesn't do much for now other than create artifacts for each tested platform from the master branch.

.github/workflows/_artifact.yml
```diff
@@ -0,0 +1,105 @@
...
+        run: |
+          OPT_INPUT="${{ inputs.opt }}"
+          make -f Makefiles/${{ matrix.makefile }} release NET_BOARD=1 OUTFILE=../supermodel${{ matrix.outfile_ext }} OPT="${OPT_INPUT:-"-O3 -g"}" LDOPT="" SDL_FRAMEWORK_RELEASE=${{ steps.find_sdl.outputs.name }} SDL_NET_FRAMEWORK_RELEASE=${{ steps.find_sdl_net.outputs.name }}
+      - name: Validate
+        run: ./supermodel${{ matrix.outfile_ext }} -print-games
+      - name: Archive
+        run: |
+          mkdir NVRAM Saves ROMs
+          cp Docs/LICENSE.txt Docs/README.txt .
+          echo "NMRAM files go here." > NVRAM/DIR.txt
```
trzy, line 93: NMRAM -> NVRAM

Makefiles/Makefile.OSX
```diff
@@ -1,84 +1,131 @@
...
-#
-# Makefile for Mac OS X systems using clang.
-#
-
-
-###############################################################################
+##
+## Supermodel
+## A Sega Model 3 Arcade Emulator.
+## Copyright 2003-2022 The Supermodel Team
```
trzy, line 4: I think you might have missed a commit where I updated the copyright header to "2003-2024" for all the Makefiles and also README.md. Would be nice to get this change in.

trzy (changes requested):
Looks like everything is in place. I just need to test this out before approving for merge. Will try to give it a spin tonight. My understanding is that I have to make a private fork and test there?

.github/workflows/_artifact.yml
h0tw1r3: fixed

Makefiles/Makefile.OSX
h0tw1r3: indeed, added.

h0tw1r3:
> Looks like everything is in place. I just need to test this out before approving for merge. Will try to give it a spin tonight. My understanding is that I have to make a private fork and test there?

No need to test on private fork. Workflows are successfully running here now.

h0tw1r3:
- Moved package creation from the github action workflow to make. Easier to test locally. Doesn't support windows command prompt. Assumed that was acceptable as the build bot script didn't either.
- Fixed a make parallel bug when using the release target

h0tw1r3:
Seems that parallel builds fail sometimes.
https://github.com/trzy/Supermodel/actions/runs/8645740818/job/23703542520#step:5:193

Disabled parallel make in workflow job for now.

trzy:
> * Moved release package creation from the github action workflow to make. Easier to test locally. Doesn't support windows command prompt. Assumed that was acceptable as the build bot script didn't either. Use `make pkg`

I don't understand why the Musashi dependency causes this problem. I've looked at the rules countless times and it seems like the dependencies are all set up correctly and make should be able to resolve them. My build bot also avoids parallel builds for this reason. I can think of a fix for this but let's not worry about it for now -- serial builds are fine.

trzy (commented):
I think this is just about ready to go. Couple nit picks/questions:

1. When this gets merged to master, the artifacts will be visible on the project page, right? If so, I'm wondering about the naming of the artifacts. When I go to "Actions" and view the builds now, they look like this:

image

Can we make these labels more consistent with the actual archive filenames (that is, including the git revision) and ensure everything is lowercased? Or am I misunderstanding how they will be displayed?

I do see in the .yml file that you are using the runner platform and arch, so if there is no built-in way to change the case, don't worry about. Maybe then change the 'supermodel-' bit to 'Supermodel-'?

2. I noticed Supermodel builds now fail on MSYS on my local machine with some linker errors involving -lmingw32, etc. Yet these work in command prompt. Strange. No need to fix this, I'll eventually look into it myself, but I don't fully understand what's going on. I might have just screwed up my environment variables at some point or something.

h0tw1r3:
Added changelog generator to `make pkg`

h0tw1r3:
> I think this is just about ready to go. Couple nit picks/questions:
> 
>     1. When this gets merged to master, the artifacts will be visible on the project page, right? If so, I'm wondering about the naming of the artifacts. When I go to "Actions" and view the builds now, they look like this:
> 

Job artifacts are simply the results of a job.

To compare it to the current build-bot process, there are separate functions/steps that build, generate a changelog, stage the release, and then upload to supermodel3.com via sftp.
What's missing from this github workflows is the equivalent of the upload (eg. publish a release).

We need to discuss how releases created. https://github.com/trzy/Supermodel/releases
It's fairly straight-forward to add a step to the workflow, but most projects use tags to trigger the creation of a new release. Right now there are no tags. Would it be possible to start using tags, ideally with [semantic versioning](https://semver.org/)?

In lieu of figuring that out, I'm trying to work out a way to publish releases that use the same 0.3a-git-<ref> format that's been historically used by the project.

> Can we make these labels more consistent with the actual archive filenames (that is, including the git revision) and ensure everything is lowercased? Or am I misunderstanding how they will be displayed?
> 

The artifact names can be changed, but they are not intended to be used for releases. If you download one, you'll see it's double archived. This is a limitation (intended feature) of the artifacts system, job artifacts are *always* wrapped in a zip file. I believe Gitlab does it the same way, in that artifacts are intended to be used by other jobs/steps in a standard way. Eg. the release step would extract the artifact, make any final changes (code signing, permissions, etc), bundle up with an installer, and then publish the release.

> I do see in the .yml file that you are using the runner platform and arch, so if there is no built-in way to change the case, don't worry about. Maybe then change the 'supermodel-' bit to 'Supermodel-'?

It's simple enough to make the final release archive file name match the current conventions.

trzy:
Sorry for my slowness, this GitHub CI/release stuff is still very unfamiliar to me, so I'm not sure what is required of the release process.

Semantic versioning is a possibility and has obvious advantages (particularly in making it easy to compare versions chronologically). 

OTOH, given the light touch administration of this project, providing builds off of git commits (and SVN commits before that) has proven to work really well for us. 

Would it be possible to do something like a weekly build (provided there was a commit) trigger?

If not, I suppose we can add tags for semantic version numbers and perhaps make the package flow create a version number from both the tag and the date? E.g.: 0.3.1-20240411 ?

How does the release process work in GitHub? Is it just a matter of a contributor running a "git tag" and then the workflow triggers? I assume there is some way to restrict permissions for who can add tags.

trzy:
I think auto-scheduled releases based on commits would be awesome. It could be like a nightly or weekly build. Would it be possible to do it once per period (day, week, whatever) and tag it with date rather than the number of commits?

PR #79, opened 2023-07-05
Set soundboard CPU to correct clock speed

gm-matthew:
The 68K on the soundboard is rated at 12 MHz but runs at 11.2896 MHz, which is 256 cycles/sample with a 44100 Hz sample rate. Removed SoundClock and Freq as they are not needed

This by itself does not fix music tempo being a little too fast


trzy:
Does this have any perceptible effect on audio? Was about to merge but haven't had time to test.

PR #81, opened 2023-08-19
DMA device register always returns Step 1.x PCI ID

gm-matthew:
Requesting PCI ID via the DMA device register on step 2.x appears to return the PCI ID of Mercury which is the ASIC connected to the PCI slot on Step 1.x, hence the PCI ID returned is the same as the Step 1.x ID (0x16c311db.

Step 2.x games by AM3 request PCI ID this way and expect to see the Step 1.x ID; up until now it was necessary to manually assign them with the Step 1.x ID to get them to run.


trzy:
Looks great! One request: there is a comment atop Real3D.cpp discussing how the PCI ID obtained via DMA register is different. Can you update that comment?

gm-matthew:
Updated the comment at the top of Real3D.cpp.

PR #82, opened 2023-08-21
Flatpak manifest

jflatt:
Manifest and supporting files for flatpak creation


flatpak/com.supermodel3.Supermodel.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - install -Dm 0755 bin/supermodel ${FLATPAK_DEST}/bin/supermodel
+      - install -Dm 0644 ${FLATPAK_ID}.desktop ${FLATPAK_DEST}/share/applications/${FLATPAK_ID}.desktop
+      - install -Dm 0644 ${FLATPAK_ID}.metainfo.xml ${FLATPAK_DEST}/share/metainfo/${FLATPAK_ID}.metainfo.xml
+      - install -Dm 0644 supermodel128x128.png ${FLATPAK_DEST}/share/icons/hicolor/128x128/apps/${FLATPAK_ID}.png
+      - cp -r Assets ${FLATPAK_DEST}/bin/
+      - cp -r Config ${FLATPAK_DEST}/bin/
+    sources:
+      - type: git
+        url: https://github.com/trzy/Supermodel.git
+        commit: b2fee4242cd061a55e8f3bb2f6dfafbf9f87403f
```
trzy, line 36: Is this tied to a particular git sha? If so, I think this is problematic. No given sha is an official release and we are constantly updating. Is there any way to have it fetch the latest? 

Secondly, is it creating the required directories: Saves/, NVRAM/, and also copying over README.txt and LICENSE.txt? I don't see these mentioned. These are required for distribution.

trzy (changes requested):

flatpak/com.supermodel3.Supermodel.yaml
jflatt: Yes, tied to a specific commit sha hash.  I suppose watching for the latest commit and triggering a flatpak update is something best handled through GitHub actions.  Although, having a specific hash allows for a little flexibility in case of breakage.

If I'm not mistaken, Saves and NVRAM directories can be created automatically by the application.  In the flatpak sandbox, and also on locked-down read-only OS, these directories are best created off of the home directory where writes are allowed, and will survive application updates.  This requires a bit of user configuration at the start.

I added the Docs directory.  There is an issue that those documentation files are not easily accessible in the flatpak filesystem.

PR #83, opened 2023-08-28
Prevent "ROLLING START" scrolling glitch in Scud Race

gm-matthew:
In Scud Race there is a glitch involving the "ROLLING START" banner. The tilemap layer that contains the banner is layer A' and it is scrolled left by 8 pixels every frame, and after exiting the left side of the screen it wraps back around to the right. Because the wraparound occurs every 512 pixels and the banner is 480 pixels long, a mask is used to hide every other appearance of the banner.

The glitch is that every four frames, the eight leftmost pixels of the banner that should be invisible manage to peek though and become visible; after much debugging I have determined that the glitch is due to the developers making the layer A' mask slightly too long on the right hand side. There seems to be some tilegen quirk that causes this glitch not to happen on real hardware, but we can't be 100% sure what it is without testing.

Rather than rewrite the tilegen code to try and replicate a quirk that we're not even sure how to reproduce, I decided it would be much easier to just patch all of the Scud ROM sets to shorten the layer A' mask and stop the glitch from occurring.

### DIFF
--- a/Config/Games.xml
+++ b/Config/Games.xml
@@ -1443,6 +1443,10 @@
       </inputs>
     </hardware>
     <roms>
+      <patches>
+        <!-- Prevent "rolling start" scrolling glitch -->
+        <patch region="crom" bits="32" offset="0x14BDB8" value="0x20B201E0" />
+      </patches>
       <region name="crom" stride="8" chunk_size="2" byte_swap="true">
         <file offset="0" name="epr-19691.20" crc32="0x83523B89" />
         <file offset="2" name="epr-19690.19" crc32="0x25F007FE" />
@@ -1531,6 +1535,8 @@
       <patches>
         <!-- Secret debug menu -->
         <patch region="crom" bits="32" offset="0x199DE8" value="0x00050208" />
+        <!-- Prevent "rolling start" scrolling glitch -->
+        <patch region="crom" bits="32" offset="0x14BDF4" value="0x20B201E0" />
       </patches>
       <region name="crom" stride="8" chunk_size="2" byte_swap="true">
         <file offset="0" name="epr-19734.20" crc32="0xBE897336" />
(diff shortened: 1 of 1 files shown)


trzy:
I'm really on the fence about this change. I always thought of ROM patches being acceptable only if strictly necessary to get the game to boot and run. This is new territory: masking emulation bugs that need solving. I'll let it in, though, because I think a lot of people will appreciate it.

dukeeeey:
The tilegen can only read 4 bytes of memory at a time? I mean to access less it has to be shifted about in registers? For instance the palette data is 50% unused with the top 16bits being completely empty. If that's true, then I wonder if the bug is related to reading the memory. For example name table is 2x 16bit words packed together, so this would cover 2 adjacent tiles. Why would you read this same memory again for the next tile? You'd just shift the data from the same memory to get the bits you need. I wonder if the bug is related to here somehow. For the wrap around, 1 tile is at on end of the tile sheet, and the other the other end, so this is the only scenario where 1 read would be impossible. You need 2 reads for the two separate memory locations.

trzy:
I do believe 16-bit access is supported. IIRC, Star Wars Trilogy does it and when it isn’t emulated you see vertical empty stripes during the vehicle info sequences that precede each level. Not sure how this is implemented on the PowerPC side. Electrically, all accesses may be the same and it might mask off only what is needed (need to check the hardware manual).

trzy:
The fact that the tile map boundary straddles a word is an interesting point. Is that where the glitch happens?

dukeeeey:
yes that's where the glitch happens. In supermodel it rolls around to the other side of the sheet, where as on the real h/w either it draws the same tile again or draws nothing. It's blank so can only guess

dukeeeey:
> Electrically, all accesses may be the same and it might mask off only what is needed (need to check the hardware manual).

yeah that is my guess how it is working looking at how the palette data is packed. I can't think why else it would be like that if it could natively read 16bits at a time without masking etc.

gm-matthew:
The tilegen RAM chip is 32-bit and so the memory is 32-bit aligned; this means that if the tilegen wants to read the 1st and 2nd tile entries from a particular row then it can read them both with a single 32-bit read, but if it wants to read the 2nd and 3rd tile entries then these come from two different 32-bit words and so it would have to perform two separate reads.

The palette RAM is 16-bit (two 8-bit chips in parallel), and is actually only 64kB even though it's mapped to 128kB from the PowerPC's perspective.

Worth noting that all tilegen RAM accesses from the PowerPC go through the tile generator itself.

PR #85, opened 2023-09-03
Ping-pong flip timing depends on the value of tilegen register 0x08

gm-matthew:
Also make the debugger display 16 bytes per line when using the "listmemory" command. Mirrored system registers can now be watched.

At some point I'd like to reimplement the RunMainBoardFrame() function but this will do for now.


gm-matthew:
Closing for now as I accidentally committed files that I changed/deleted in my personal repository

PR #101, opened 2023-11-03
Add LOD switching and blending out culling nodes

gm-matthew:
We're not blending between LODs yet as we're not sure of the best way to implement it.
Also implementing control bits for disable culling and enable model scale (needed for ECA and Harley)

### DIFF
--- a/Src/Graphics/New3D/Model.cpp
+++ b/Src/Graphics/New3D/Model.cpp
@@ -4,12 +4,13 @@ namespace New3D {
 
 NodeAttributes::NodeAttributes()
 {
-	currentTexOffsetX	= 0;
-	currentTexOffsetY	= 0;
-	currentPage			= 0;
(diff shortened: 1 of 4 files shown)


dukeeeey:
Looks great :)

Only thing I'd change is

LODFeatureType lodTableEntry = m_LODBlendTable->table[lodTablePointer];

to

LODFeatureType& lodTableEntry = m_LODBlendTable->table[lodTablePointer];

Better to get a reference than a copy of the object. But a very minor optimisation !

trzy:
Could the scale fix be added to the legacy engine? If it just overrides transform scale, should be pretty easy.

I eventually want to remove the legacy engine but some people still use it.

dukeeeey:
The node model scale is used in the lighting calculation. If you increase the scale it makes the normals bigger so the object becomes brighter. The model scale fixes that.

PR #114, opened 2023-12-02
fixed some `-Wunused-*` compiler warnings

firewave:


trzy:
Thanks!

trzy:
No idea. They look like some sort of attempt to either work around a compiler bug or provide a breakpoint target. I can't think of a legitimate reason for them to be there but I've asked the person who I believe introduced them.

PR #115, opened 2023-12-02
fixed `-Wformat-overflow` compiler warnings / added missing newline to log messages

firewave:
Some examples for reference:
```
Src/OSD/SDL/Main.cpp:1086:30: warning: ‘ (Paused)’ directive writing 9 bytes into a region of size between 1 and 128 [-Wformat-overflow=]
 1086 |         sprintf(titleStr, "%s (Paused)", baseTitleStr);
      |                              ^~~~~~~~~
Src/OSD/SDL/Main.cpp:1086:16: note: ‘sprintf’ output between 10 and 137 bytes into a destination of size 128
 1086 |         sprintf(titleStr, "%s (Paused)", baseTitleStr);
      |         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

```
Src/OSD/Logger.cpp:382:29: warning: ‘%s’ directive writing up to 4095 bytes into a region of size 4088 [-Wformat-overflow=]
  382 |   sprintf(string2, "[Error] %s\n", string1);
      |                             ^~     ~~~~~~~
Src/OSD/Logger.cpp:382:10: note: ‘sprintf’ output between 10 and 4105 bytes into a destination of size 4096
  382 |   sprintf(string2, "[Error] %s\n", string1);
      |   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/Src/OSD/Logger.cpp
+++ b/Src/OSD/Logger.cpp
@@ -239,8 +239,8 @@ void CFileLogger::DebugLog(const char *fmt, va_list vl)
   char string1[4096];
   char string2[4096];
 
-  vsprintf(string1, fmt, vl);
-  sprintf(string2, "[Debug] %s", string1);
+  vsnprintf(string1, sizeof(string1), fmt, vl);
+  snprintf(string2, sizeof(string2), "[Debug] %s", string1);
 
   // Debug logging is so copious that we don't bother to guarantee it is saved
   std::unique_lock<std::mutex> lock(m_mtx);
@@ -257,8 +257,8 @@ void CFileLogger::InfoLog(const char *fmt, va_list vl)
   char string1[4096];
   char string2[4096];
 
-  vsprintf(string1, fmt, vl);
-  sprintf(string2, "[Info]  %s\n", string1);
+  vsnprintf(string1, sizeof(string1), fmt, vl);
+  snprintf(string2, sizeof(string2), "[Info]  %s\n", string1);
 
   // Write to file, close, and reopen to ensure it was saved
   std::unique_lock<std::mutex> lock(m_mtx);
@@ -276,8 +276,8 @@ void CFileLogger::ErrorLog(const char *fmt, va_list vl)
   char string1[4096];
   char string2[4096];
 
-  vsprintf(string1, fmt, vl);
-  sprintf(string2, "[Error] %s\n", string1);
+  vsnprintf(string1, sizeof(string1), fmt, vl);
+  snprintf(string2, sizeof(string2), "[Error] %s\n", string1);
 
   // Write to file, close, and reopen to ensure it was saved
   std::unique_lock<std::mutex> lock(m_mtx);
@@ -347,8 +347,8 @@ void CSystemLogger::DebugLog(const char *fmt, va_list vl)
   char string1[4096];
   char string2[4096];
 
-  vsprintf(string1, fmt, vl);
-  sprintf(string2, "[Debug] %s", string1);
+  vsnprintf(string1, sizeof(string1), fmt, vl);
+  snprintf(string2, sizeof(string2), "[Debug] %s", string1);
 
 #ifdef _WIN32
   OutputDebugString(string2);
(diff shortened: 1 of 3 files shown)


Src/OSD/Logger.cpp
```diff
@@ -228,10 +228,10 @@ void CFileLogger::DebugLog(const char *fmt, va_list vl)
   }
 
   char string1[4096];
-  char string2[4096];
+  char string2[4105]; // add 8 for the prefix and 1 for newline
```
trzy, line 231: These are not needed. 4096 was chosen somewhat arbitrarily. We don't actually check the size of the resulting output. If you want to fix this, you should instead prefer to use vsnprintf() and pass 4096 as the "n" parameter. Likewise, sprintf -> snprintf.

Src/OSD/Logger.cpp
```diff
@@ -228,10 +228,10 @@ void CFileLogger::DebugLog(const char *fmt, va_list vl)
   }
 
   char string1[4096];
-  char string2[4096];
+  char string2[4105]; // add 8 for the prefix and 1 for newline
 
   vsprintf(string1, fmt, vl);
-  sprintf(string2, "[Debug] %s", string1);
+  sprintf(string2, "[Debug] %s\n", string1);
```
trzy, line 234: This newline is actually incorrect. DebugLog is special and you'll notice wherever it is used in Supermodel, a \n is manually included. The intent behind this was to allow formatting debug output more freely with multiple calls to generate a single line if needed. However, now that the [Debug] tag has been added, it may be worthwhile to automatically include a \n but if this is done, all places where DebugLog is used must be updated to drop their \n.

Src/OSD/Logger.cpp
```diff
@@ -336,10 +336,10 @@ void CSystemLogger::DebugLog(const char *fmt, va_list vl)
...
```
trzy, line 342: Same here.

Src/OSD/SDL/Main.cpp
```diff
@@ -917,7 +917,7 @@ int Supermodel(const Game &game, ROMSet *rom_set, IEmulator *Model3, CInputs *In
 
   // Set the video mode
   char baseTitleStr[128];
-  char titleStr[128];
+  char titleStr[256];
```
trzy, line 920: This change is fine, I suppose. Does anything actually exceed 128 bytes? We should probably also move to using snprintf() for safety here.

trzy (changes requested):
Please see the attached comments. I don't think the buffer lengths should be touched in Logger.cpp, there is no need for it. Rather, use the safe forms of vsprintf and sprintf (vsnprintf and snprintf). Also, the \n in debug logging was intentionally omitted. It's not a typo.

trzy:
Will you be able to make the suggested changes?

firewave:
> Will you be able to make the suggested changes?

Yes. Hopefully I finally get to it this week.

Src/OSD/Logger.cpp
firewave: Done.

Src/OSD/Logger.cpp
firewave: Done.

Src/OSD/SDL/Main.cpp
firewave: As it appends up to 128 bytes from `baseTitleStr` in that string it needs to be at least `128 + n`.

```
Src/OSD/SDL/Main.cpp:1086:30: warning: ‘ (Paused)’ directive writing 9 bytes into a region of size between 1 and 128 [-Wformat-overflow=]
 1086 |         sprintf(titleStr, "%s (Paused)", baseTitleStr);
      |                              ^~~~~~~~~
Src/OSD/SDL/Main.cpp:1086:16: note: ‘sprintf’ output between 10 and 137 bytes into a destination of size 128
 1086 |         sprintf(titleStr, "%s (Paused)", baseTitleStr);
      |         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Src/OSD/Logger.cpp
firewave: Okay. I moved back to the original buffer length and use `*snprintf()` now.

firewave:
There are additional warnings now but I cannot change that file because there are some issues with the line endings.

```
Src/Debugger/CPU/68KDebug.cpp:1353:56: note: ‘sprintf’ output between 15 and 64 bytes into a destination of size 50
 1353 |                                                 sprintf(moveStr, "movem.%c %s,[AM2]", sizeC, regList);
      |                                                 ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

trzy:
Closed because no longer needed (the warnings have been fixed).

PR #124, opened 2023-12-25
Reset frustum plane correction factor for viewports that do not fill screen

gm-matthew:


dukeeeey:
already fixed in the source :)

PR #129, opened 2023-12-30
No need to have changelog from 2011 in Readme.txt anymore.

jaoxford:


jaoxford:
Thought it would make sense to trim down the readme file to remove the changelog from 2011.

trzy:
The changelog that is automatically generated does not contain these entries for past releases. But it does make sense to remove them for now as long as we are not cutting official releases anymore.

PR #136, opened 2024-03-05
fix segfault in CMultiLogger

baraclese:
Tested on linux with glibc.

A segfault can be triggered when Supermodel does not find a Supermodel.ini file to load. At that point it will call CConsoleErrorLogger::ErrorLog and then CFileLogger::ErrorLog.

The reason for the segfault is that vsprintf may change its va_list argument so repeatedly calling it with the same va_list arg is undefined behavior.

Fix this by creating a copy of the va_list argument before each vsprintf call.


trzy (approved):

trzy:
Thank you!

PR #139, opened 2024-03-06
option to specify game in multi-romset

h0tw1r3:
new optional argument `-game` is useful for selecting a specific game from a romset zip file that contains multiple versions.

### DIFF
--- a/Src/GameLoader.cpp
+++ b/Src/GameLoader.cpp
@@ -663,7 +663,6 @@ void GameLoader::IdentifyGamesInZipArchive(
 
 void GameLoader::ChooseGameInZipArchive(std::string *chosen_game, bool *missing_parent_roms, const ZipArchive &zip, const std::string &zipfilename) const
 {
-  chosen_game->clear();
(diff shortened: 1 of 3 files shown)


trzy:
Will look at this PR this weekend. Thank you! This is a band-aid solution as far as I'm concerned. I'd like to rewrite the ROM loading system to:

1. Continue to use load-by-CRC rather than filenames. Why? Because I still use ancient ROMs that are just fine except that they don't conform to the latest MAME naming convention, whatever that may be today. I suspect many are actually in this situation and have ROM sets that work fine.

2. Allow multiple "dependent" ROMs to be specified (we currently allow one "parent"). This might need to be an entire DAG of dependencies given how things seem to be structured in MAME now.

This is actually not that conceptually difficult but it is a lot of sweat and toil. I might try to prototype this in Python first and then port it to C++ someday. I think the initial ROM set and game should be detected from the file passed and then any dependencies should be loaded from the same directory assuming the zip file is named "game.zip".

PR #143, opened 2024-03-14
support engine selection with print-gl-info argument

h0tw1r3:
print-gl-info always defaults to legacy, which on mac at least, returns unexpected info.
with this PR the default engine is now new3d, or whatever is globally configured.

Before (only lines that changed):

```
GPU info: 2.1 Metal - 88
  Version                  : 2.1 Metal - 88
  Shading Language Version : 1.20
```

After

```
GPU info: 4.1 Metal - 88 (core profile)
  Version                  : 4.1 Metal - 88
  Shading Language Version : 4.10
```

Also supports specifying the engine and printing info, eg `supermodel -legacy3d -print-gl-info`

### DIFF
--- a/Src/OSD/SDL/Main.cpp
+++ b/Src/OSD/SDL/Main.cpp
@@ -378,7 +378,7 @@ static void PrintGLInfo(bool createScreen, bool infoLog, bool printExtensions)
   unsigned xOffset, yOffset, xRes=496, yRes=384, totalXRes, totalYRes;
   if (createScreen)
   {
-    if (OKAY != CreateGLScreen(false, false, "Supermodel - Querying OpenGL Information...", false, &xOffset, &yOffset, &xRes, &yRes, &totalXRes, &totalYRes, false, false))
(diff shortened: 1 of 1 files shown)


dukeeeey:
for gpu info maybe more useful to get the max supported opengl version? This works on windows ... just not on apple because it doesn't support compatibility profiles.

h0tw1r3:
> for gpu info maybe more useful to get the max supported opengl version? This works on windows ... just not on apple because it doesn't support compatibility profiles.


I think the expectation is to show what is initialized by SDL. This PR simply makes it accurate to selected engine.

trzy:
Is there a way to determine the maximum GL version on macOS?

I can see this being useful but if the legacy engine is broken on macOS, seems moot and we should do away with -legacy3d there. Then on Windows and Linux this becomes somewhat irrelevant?

dukeeeey:
I don't think you can, you just have to create a context and see if it fails or not. But the tilegen I rewrote for opengl 3+ because it uses unsigned integer maths. It can't work with a gl 2.1 context. The only way to make it work on apple is to port the legacy renderer to opengl 4 and create a 4.1 context. Really not worth the hassle unless the performance of the new renderer is THAT bad under apple.

trzy:
Closing this. I cherry-picked this change into a PR I'm about to commit that also removes the legacy renderer entirely from macOS. Thanks, @h0tw1r3 for doing this. I think it makes sense.

PR #147, opened 2024-05-02
Adding newly dumped version of Scud Race

gm-matthew:
Allow Step 1.5 games to access 53C810 via 0xC00000xx if the netboard is disabled; scuddxo requires this for 3D graphics to work

Also scuddx does not use the netboard


gm-matthew:
LoadGame() has a line of code that converts the game stepping string to an integer value before passing it to GPU.SetStepping(); perhaps we could save this value and compare using it instead of the string. I’ll have a crack at it this evening.

PR #162, opened 2024-07-14
Give PowerPC more cycles to acknowledge sound IRQs

gm-matthew:
400 cycles was not enough and would cause VF3 to hang. 1000 cycles should be more than plenty


trzy (approved):
1000 cycles. Damn. Do they have Taka-Arashi waddling around in there taking care of int acks??

gm-matthew:
To be honest 500 cycles would probably have been enough, but I thought I'd bump it up to 1000 just to be extra safe :-)

PR #167, opened 2024-07-23
Scale texture NP values for luminous polygons even if modelScale is not set

gm-matthew:
My implementation of texture NP has resulted in a few cases of scaled up polygons (that are not using the modelScale parameter) using lower-quality mipmaps than they should; these include the lens flare effect in Scud Race and Daytona 2 (possibly also Virtua Striker 2), the fish in Sega Bass Fishing and the map logo and window background in Harley.

All of these cases are luminous polygons, so I've come up with this modification that fixes all of the cases I've managed to find without introducing excessive aliasing to scaled up polygons that currently look fine, e.g. dot matrix display from Expert course in Dirt Devils, Death Star during cutscene in Star Wars Trilogy.

### DIFF
--- a/Src/Graphics/New3D/R3DShaderQuads.h
+++ b/Src/Graphics/New3D/R3DShaderQuads.h
@@ -11,6 +11,7 @@ uniform float	nodeAlpha;
 uniform mat4	modelMat;
 uniform mat4	projMat;
 uniform bool	translatorMap;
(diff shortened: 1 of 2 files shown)


gm-matthew:
Closing this pull request because I believe I have discovered the key to implementing a better solution.

PR #173, opened 2024-08-01
New textureNP implementation

gm-matthew:
This new implementation uses the previously unused viewport parameter "cota".

At the moment it does not compensate for higher resolutions, partly because doing so would result in microtextures appearing sooner than they should - which would be particularly undesirable for L.A. Machineguns because it has microtextures that aren't even supposed to be there - and most of the textures already look pretty sharp even at 1080p.

### DIFF
--- a/Src/Graphics/New3D/Model.h
+++ b/Src/Graphics/New3D/Model.h
@@ -200,6 +200,7 @@ struct Viewport
 	float	angle_right;
 	float	angle_top;
 	float	angle_bottom;
+	float	cota;
 
 	Mat4	projectionMatrix;		// projection matrix, we will calc this later when we have scene near/far vals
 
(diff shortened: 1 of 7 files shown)


trzy:
Out of curiosity, what do you mean by compensation for higher resolutions? I typically run at higher than 1080p, for example.Sent from my iPhoneOn Aug 2, 2024, at 12:03 AM, dukeeeey ***@***.***> wrote:﻿
Merged #173 into master.

dukeeeey:
The cota viewport value when it's calculated takes into consideration the screen resolution and clipping planes. I think it provides something like a poly to pixel mapping for different depths in the scene.

Mipmapping on the model 3 looks to be a crude approximation of what modern hw does. It's fairly ugly in a lot of scenarios lol

Might provide a fallback for less accurate but more beautiful mipmaps.

PR #179, opened 2024-08-09
Added support for custom MPEG music in a new Music.xml config file

trzy:
Someone in the past had made a patch to Supermodel that allowed custom MPEG music to be substituted in. I had a spare moment tonight and saw a request for this on the forums, so decided to give it a go. MP3 files (and I assume MP2) can be specified in Config/Music.xml now. 

The byte offsets of tracks must be known. There is also support for specifying a byte offset within the file (for looping) but this remains a "hidden" option for now. 

At the moment, only the addresses of Scud Race's tunes are given. There is no mechanism for discovering these without inserting debug printf() statements for now. I will come back at a later time and try to add more examples for popular games like Daytona 2.

### DIFF
--- a/Config/Music.xml
+++ b/Config/Music.xml
@@ -1,0 +1,31 @@
+<!--
+  Supermodel
+  A Sega Model 3 Arcade Emulator.
+  Copyright 2003-2024 The Supermodel Team
+
+  Music.xml
+
+  This file defines custom MPEG music tracks to be used in DSB1 and DSB2 games.
+  Make sure to use MP3 files sampled at 32 KHz. Normal sampling rates (e.g. 44.1
+  or 48 KHz) will sound slow.
+
+  The example below is wrapped in a dummy <comment></comment> block to render it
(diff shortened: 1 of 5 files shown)


dukeeeey:
I mean, I personally think these sort of changes would be better off in a branch or something. For the simple reason if it doesn't work or has bugs someone has to maintain it. But it's up to you, maybe people would be crazy for this feature.

trzy:
Accurate emulation should always take priority over bells and whistles like this, and I understand the maintenance concern, but this one is a fairly trivial change and could easily be removed if a change to the MPEG system was needed. 

It would be interesting to have branches with different features but not sure how that would affect the release system. Could be a mess.

I think ultimately the best solution would be to add scripting language hooks. This was demonstrated by someone using Lua although I’ve been hoping a nicer language than Lua would show up while still being as easy to integrate.

I’m going to reply to the Dev plan thread on the forums later today with some thoughts on a few other features I wanted to slowly add to Supermodel.

PR #182, opened 2024-08-14
New JTAG implementation

gm-matthew:
Sega Rally 2 and Star Wars Trilogy no longer require patches to run; these patches have been removed from Games.xml (although the games will still function just fine with the patches left intact).

Step 2.x games now display "daughter board was detected" or similar as they do on real hardware.

The savestate file version has been updated so old savestates will not work; we may revert this before committing to provide partial support for old savestates.

We do not yet implement the white flash which is triggered by setting the "block culling" bit in the Mercury modeword; this should be easy to do.


dukeeeey:
char* str = new char[MAX_REGISTER_LENGTH + 1];
This leaks? There is no free? :)
You could use
std::unique_ptr<char[]> str(new char[MAX_REGISTER_LENGTH + 1]);
or just call
delete [] str;

Also I'd personally avoid the global MAX_REGISTER_LENGTH
I'd make it a public static const variable and stuff it in the CJTAGDevice class.
Less likely to have collisions with global namespace then

But otherwise looks great :)

dukeeeey:
m_shiftRegSize
This is a size_t variable, which means it'll be different sizes depending on platform, ie if people build a 32build for example. These saves states will break between versions. This is a very bad idea lol

uint32_t is enough. Size_t is 64 bit under win64.

I personally try and avoid these variable length variables unless you really need them (ie to store a pointer or something)

gm-matthew:
Fair point about size_t, forgot that 32-bit builds are still a thing. Will change it to uint32_t

Src/Model3/JTAG.cpp
```diff
@@ -35,186 +35,337 @@
...
+        m_jupiter.SetIDCode(0x116c7057);
+        m_numDevices = 10;
+    }
+}
+
+CJTAG::CJTAG(CReal3D& real3D) :
+    m_mercury(real3D, CASIC::Name::Mercury),
+    m_venus(real3D, CASIC::Name::Venus),
+    m_earth0(real3D, CASIC::Name::Earth),
+    m_earth1(real3D, CASIC::Name::Dummy),
```
trzy, line 363: These dummy devices are just shadows of the same device right? Do we know how this works? I assumed that the JTAG shift registers are physically located inside each ASIC. How can there be copies like this? Is it possible instead that the shift registers inside these ASICs are actually 2x or 4x (in the case of Mars) as big, and internally the same data is replicated N times? 

That is, could this also be handled by having only a single ASIC but then increasing the internal shift register size and replicating the ID code, mode word, etc.? This would complicate the interface (e.g., you'd need an extra parameter in the constructors to indicate how many times to replicate -- 2 or 4.

Perhaps a comment at the top of the file describing what's going on?

Src/Model3/JTAG.cpp
```diff
@@ -35,186 +35,337 @@
...
+    m_shiftReg = 1;
+}
+
+void CASIC::UpdateDR()
+{
+    switch (m_instructionReg)
+    {
+    case 8:                     // MODEWORD
+        m_modeword = m_shiftReg.to_ulong();
+        m_real3D.WriteJTAGModeword(m_deviceName, m_modeword);
```
trzy, line 112: What is the "mode word"? I assume this is a Real3D specific thing that they put in all their ASICs? Would it be possible to elaborate on what this is and how it was discovered (especially if there is any documentation to cross-reference; I can't remember if it appears in the SDK or not) in a descriptive comment at the top of one of the files?

Src/Model3/JTAG.h
```diff
@@ -29,60 +29,141 @@
...
+    CASIC m_mercury;
+    CASIC m_venus;
+    CASIC m_earth0, m_earth1;
+    CASIC m_mars00, m_mars01, m_mars10, m_mars11;
+    CASIC m_jupiter;
+    C3DRAM m_3dram[8];
+
+    CJTAGDevice* m_device[17] =
+    {
+        &m_jupiter,
```
trzy, line 148: Although it is pretty obvious from reading the code that these should not be reordered it might make sense to add a one-line comment to the effect of "Do not reorder. This is the order in which devices appear in the bit stream."

Src/OSD/SDL/Main.cpp
```diff
@@ -681,7 +681,7 @@ static void TestPolygonHeaderBits(IEmulator *Emu)
  Different subsystems output their own blocks.
 ******************************************************************************/
 
-static const int STATE_FILE_VERSION = 3;  // save state file version
+static const int STATE_FILE_VERSION = 4;  // save state file version
```
trzy, line 684: This is good -- I think trying to back out fragments of the JTAG state (like maybe the modewords) from existing save states would be a huge mess. I'm good with the save state version being bumped forward.

trzy (changes requested):
This is awesome! Thank you for putting in the effort to do this. My only real suggested changes are to add a paragraph or two of explanation on your findings at the top of one of the files, wherever you deem appropriate. The logic is very readable but I think it's good to have a bird's eye overview, especially if it's possible to cross-reference any documentation that we have (i.e., the SDK header files, JTAG spec, etc.) if only to jog our memories in the future if something needs to be changed (e.g., I think the VROM test is done via JTAG boundary scan and could someday be emulated here).

trzy:
Oh -- and one last thing: my old BitRegister data structure won't be necessary anymore. You can go ahead and remove it from the project.

Src/Model3/JTAG.cpp
gm-matthew: The mode word appears prominently in the Real3D firmware which is how I discovered it; it certainly does appear to be a Real3D specific thing that all the ASICs use.

I made a post on the old forum that lists all of the modeword bits for each ASIC as described in the firmware: https://supermodel3.com/ArchivedForum/192.168.0.100/Forum/viewtopic47b9.html?f=7&t=2282&sid=cef2229983e7ae61b43df8e35d27077f

The sun-clamp disable bit in the Mars ASIC is listed as unused, but I'm guessing that it was added in a later revision because it is definitely used by both Daytona 2 PE and L.A. Machineguns.

Src/Model3/JTAG.cpp
gm-matthew: If we take the Mars ASIC for example, there are two of them on step 1,x and four of them on step 2.x. Each of them has their own shift register, modeword, etc. In theory you could disable sun clamping for only one Mars ASIC and it would only affect every other pixel on step 1.x and additionally only on odd/even scanlines on step 2.x.

This would be difficult to emulate accurately since we just use OpenGL to render the whole thing, so I made it so that only the first of each ASIC is hooked up to the CReal3D class; the dummy device just tells it to do nothing.

dukeeeey:
Dumb question maybe about jtag. Aren't they all connected together? I thought the output of 1 connected to the input of another. So a sort of giant shift register.

gm-matthew:
> Dumb question maybe about jtag. Aren't they all connected together? I thought the output of 1 connected to the input of another. So a sort of giant shift register.

They are all connected together so they do effectively act like one giant shift register, which is how the old implementation treated it. However, this "giant shift register" can change size depending on the instruction - the old implementation was fixed at 197 bits, which worked for the IDCODE instruction applied to all ASICs but not really for anything else.

trzy (approved):

PR #188, opened 2024-09-04
slightly optimize depth & stencil usage

toxieainc:
use only a single glReadPixels for LOS
avoid creating a depth/stencil buffer in the New3D mode
wire error handling for CreateFBO(DepthCopy)
handle black FB init for triple buffering

### DIFF
--- a/Src/Graphics/Legacy3D/Legacy3D.h
+++ b/Src/Graphics/Legacy3D/Legacy3D.h
@@ -231,7 +231,7 @@ struct TexSheet
  *
  * 3D renderer. Lots of work to do here :)
  */
-class CLegacy3D: public IRender3D
+class CLegacy3D final : public IRender3D
 {
 	friend class CTextureRefs;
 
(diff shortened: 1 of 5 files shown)


dukeeeey:
return ((CreateFBODepthCopy(width, height) == OKAY) && (fboStatus == GL_FRAMEBUFFER_COMPLETE)) ? OKAY : FAIL;

The OKAY and FAIL things are really not needed. The return type is a boolean. I know they are used in other parts of the code but I think it's poor design. The defines are flipped from what you might expect.

toxieainc:
Should i rather do a follow up commit that then removes all of the OKAY/FAIL stuff all over the project?

dukeeeey:
> Should i rather do a follow up commit/PR that then removes all of the OKAY/FAIL stuff all over the project?

Yeah if you want to make a pr for this that's fine.

toxieainc:
By now its used through this PR.. (unless i forgot something)

PR #192, opened 2024-09-13
Fix -print-gl-info after fb785b36bd70c16001336e22a75950ba5ef3808f

DirtBagXon:
Fixes:

```
Copyright 2003-2024 by The Supermodel Team
terminate called after throwing an instance of 'std::range_error'
  what():  Node "New3DEngine" does not exist
Aborted (core dumped)
```

### DIFF
--- a/Src/OSD/SDL/Main.cpp
+++ b/Src/OSD/SDL/Main.cpp
@@ -278,7 +278,7 @@ static Result CreateGLScreen(bool coreContext, bool quadRendering, const std::st
   SDL_GL_SetAttribute(SDL_GL_RED_SIZE,8);
   SDL_GL_SetAttribute(SDL_GL_GREEN_SIZE,8);
   SDL_GL_SetAttribute(SDL_GL_BLUE_SIZE,8);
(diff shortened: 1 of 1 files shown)


trzy (changes requested):
In this case, I think the correct fix is actually to check for ValueAsDefault<bool>(true). Note that this is what the rest of the function does for other values. The default value for New3DEngine is true these days. 

Would also be good to make sure other ValueAsDefault checks in the same function match the actual defaults set elsewhere in Main.cpp.

ValueAs<>() will throw an exception if a key does not exist in the config. ValueAsDefault<>(DEFAULT) will return DEFAULT if the value does not exist.

DirtBagXon:
I run a fork where the default may not be `New3DEngine` - but if `ValueAsDefault` works here, changed.

It fixes the segfault in any case.

Thanks.

trzy (approved):
The best fix would probably be to run this after the config is properly parsed but that would require rearranging how things happen in main() so that when this option is detected, no file is required. It’s a bit more work and we can do it in the future.

PR #193, opened 2024-09-18
Fix texel offsets when sampling the AA buffer

toxieainc:
..otherwise one can see artifacts and offsets, e.g. when rendering via the default 'original size' windowed mode and comparing to no AA.

There is still some tiny offset visible when comparing no AA vs multiple samples on higher resolutions, but as this offset differs between 2D and 3D screen elements, it is rather related to the resolution scaling itself, not the AA sum up pass.


dukeeeey:
yeah I think it should have be texSize-1 instead of just texSize for texel fetch

PR #194, opened 2024-09-18
Increase precision and performance for Base/Alpha Layer compositing

toxieainc:
Not as bad as #193, but still worth it


dukeeeey:
Why have you done this
/*-vec2(0.5)*/
In the code
The normal solution to convert to normalised coordinates to texel coordinates you just multiply them by width-1 and height-1. I think that is a much cleaner solution and gives the expected result.

toxieainc:
Thanks!

The -vec2(0.5) is there, as this would be what we wanted if we would work with floating point texel reads (interpolated), but as we are interested in the raw pixels only and use texel fetch with integer coords, better not screw around with potential fp32 issues.

And yes, the other way should also work, but similar to what is happening in Render2D (=a weird offset is needed in there), there is still something off, so using the fragment coordinates will always guarantee to have a perfect 1:1 match (and no full texture pipeline is triggered, only a single texel read, which can be faster, especially on lowest end GPUs).

Issue #189, opened 2024-09-10
Gamma


toxieainc:
Hi!

This is not so much an issue, but more like a question, or comment on the rendering pipeline:
As e.g. the Virtua Racing port on the Switch stumbled over this, i wonder how gamma correction is being done in Supermodel.
In VR, there was initially an issue where the gamma was not compensated for the ArcadeCRT<->LCD gamma differences, thus colors were very off. They patched this later on and offered a gamma slider (not the best solution, but better than ignoring it).

So how is this done in Supermodel? Cause maybe that could help to further improve visual fidelity/precision, depending on how/if the gamma compensation is currently done implicitly during rendering itself, instead of a postprocess (which then affects the final pixels instead of parts that are still composited, etc).

dukeeeey:
There is no gamma correction. I know the original real3d pro-1000 had gamma correction functionality, but as far as we know it's not used on the model 3. I assume image we draw if rendered on an original screen would look the same, there for if you wanted to emulate a CRT you would need some sort of post process filter.

toxieainc:
Okay, this is what i would have assumed as well.
So basically all lighting/material computations are done in whatever the textures/color-parameters color space are in, and then that is as-is composited and blitted to the screen.
That means that (in theory) one should at least compensate for the CRT vs LCD gamma difference in an optional(!) pass.
I will experiment with that then, thanks!

toxieainc:
Do you guys know which monitors were used in the Model 3 machines?

From my experiments so far, using a color profile that matches common/higher-end japanese CRTs of the time, i think i get closer to the colors that can be seen on youtube vids (e.g. the roads in Daytona 2 seem to be more blue in the videos).

trzy:
Unfortunately I don’t. I don’t think anyone ever produced a good description of the non-PCB components :(Sent from my iPhoneOn Sep 16, 2024, at 2:09 AM, toxie ***@***.***> wrote:﻿
Do you guys know which monitors were used in the Model 3 machines?
From my experiments so far, using a color profile that matches common/higher-end japanese CRTs of the time, i think i get closer to the colors that can be seen on youtube vids (e.g. the roads in Daytona 2 seem to be more blue in the videos).

p1pkin:
> Do you guys know which monitors were used in the Model 3 machines?
> 
[...]

PR #195, opened 2024-09-20
Implement CRT color emulation

toxieainc:
so not scanlines or other CRT aspects, just the differences in the region-specific TV color standards of the time. These came with the caveat that displaying video content on other regions TVs would/should have needed color corrections (which was usually not done for video games, only broadcasting, etc). Nowadays the world has settled on the same standards and we can transform from the old ones to the (for now) common sRGB.

This means one can choose between 6 different mappings now:
0=none (as before/default)
1=ARI/D93 (recommended for all JP developed games)
2=PVM_20M2U/D93
3=BT601_525/D93
4=BT601_525/D65 (recommended for all US developed games)
5=BT601_625/D65 (recommended for all EUR/AUS developed games)

wired into the supersampling summation pass, as this is the final output pass

addresses #189

### DIFF
--- a/Config/Supermodel.ini
+++ b/Config/Supermodel.ini
@@ -37,6 +37,11 @@ QuadRendering = false
 WideScreen = false
 Stretch = false
 WideBackground = false
+Supersampling = 1
+; CRT-like color adaption: 0=none, 1=ARI/D93 (recommended for all JP developed games),
+; 2=PVM_20M2U/D93, 3=BT601_525/D93, 4=BT601_525/D65 (recommended for all US developed games)
+; 5=BT601_625/D65 (recommended for all EUR/AUS developed games)
+CRTcolors = 0
 
 ; Refresh rate (milliHertz accuracy). Actual Model 3 refresh rate is 57.524 Hz
 ; but this can cause judder so we use 60 Hz by default.
--- a/Docs/README.txt
+++ b/Docs/README.txt
@@ -1213,6 +1213,25 @@ not.  All options are case sensitive.
 
     ----------------
 
+    Option:         -crtcolors=<mode>
+
+    Description:    Emulates other regional TV standards of the time.
+                    These came with the caveat that displaying video content on
+                    other regions TVs would/should have needed color corrections
+                    (which was usually not done for video games, only broadcasting, etc).
+                    Nowadays the world has settled on the same standards, and we
+                    can transform from the old ones to the (for now) common sRGB.
+
+                    This means one can choose between 6 different mappings now:
+                    0=none (default)
+                    1=ARI/D93 (recommended for all JP developed games, most likely true for all Model 3 games)
+                    2=PVM_20M2U/D93
+                    3=BT601_525/D93
+                    4=BT601_525/D65 (recommended for all US developed games)
+                    5=BT601_625/D65 (recommended for all EUR/AUS developed games)
+
+    ----------------
+
     Option:         -show-fps
 
     Description:    Shows the frame rate in the window title bar.
--- a/Src/Graphics/SuperAA.cpp
+++ b/Src/Graphics/SuperAA.cpp
@@ -1,14 +1,14 @@
 #include "SuperAA.h"
 #include <string>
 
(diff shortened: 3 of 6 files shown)


toxieainc:
..interesting games to try with mode 1: Sega Rally 2, Daytona 2, Scud Racing.

dukeeeey:
A few comments. I'd use a enum instead of int CRTcolors since it's immediately clear what this is, and what the range of possible values are. I know it would have to be cast to an int to pass to a shader but that's fine. GLSL doesn't support enums but you can use some #DEFINES instead of random numbers.

This is almost unreadable

			//return pow(color, 1.0/2.2);
			//if (!(color > 0.0)) // also covers NaNs
			//	return 0.0;
			/*else*/ if (color <= 0.0031308)
				return 12.92 * color;
			else //if (color < 1.0)
				return 1.055 * pow(color, 1.0/2.4) - 0.055;
			//else
			//	return 1.0;

Please remove code that is effectively dead.

Maybe this code

		std::string ccString = "const int cc = ";
		ccString += std::to_string(m_crtcolors);
		ccString += ";\n";
		std::string cgString = "const float cgamma = ";
		cgString += (m_crtcolors == 1) ? "2.5" : (m_crtcolors == 2) ? "2.25" : "-1.0"; // what is the 'real' gamma/EOTF of a japanese arcade CRT of that time? 2.4 or 2.5 is what the net so far agrees on
		cgString += ";\nconst mat3 colmatrix = ";
		cgString += (m_crtcolors == 1) ? "colARI" : (m_crtcolors == 2) ? "colPVM" : (m_crtcolors == 3) ? "colBT601JP" : (m_crtcolors == 4) ? "colBT601US" : "colBT601EA";
		cgString += ";\n";

would be way more readable done as #ifdef inside the shader itself. Where you just pass the CRT colour value.

So something like

#ifdef CRTCOLOURS==0
  const float cgamma = 2.5;
#elif CRTCOLOURS== 1
 const float cgamma = -1;
#endif

etc

Nothing technically wrong with the code you've written but I think the latter code would be a lot easier to maintain.

toxieainc:
Okay, sounds good, thanks!

p1pkin:
> 1=ARI/D93 (recommended for all JP developed games)

I'd bet all or nearly all Model 3 games was developed in Japan. if so, why to have other options?
besides of that, what makes you think gamma because of CRT ? for example it be because of video DAC. have you checked videos captured directly from Model 3 PCB ?

toxieainc:
Why not have the other options as well? Its not like this would cost anything..
As one example one could argue that the japanses devs worked on something like a Sony PVM to tweak their art and then on the arcade monitor itself it just looked 'similar enough' for them.

As for gamma: Each TV,etc-standard that is not working in raw linear space features something that is called an EOTF, something that is commonly referred to as 'gamma', if one is not super-deep into the whole color wizardry. So this has to be compensated for. On top, there are certain quirks that are hardware dependent, and apparently arcade monitors of that era seem to feature a gamma function of 2.4-2.5. So if one wants to transform to linear space, one has to compensate for that. Then when going back to sRGB to display on 'our' common displays of today one has to do something similar (e.g. a gamma of 2.2, or the piecewise variant).

As for video captures: This is the tricky part. If one looks at hardware captures (i guess directly from the DAC), then obviously all is well, and Supermodel matches pretty much what was captured.
But this is exactly the point of this PR: Its not about the output of the board, but what was/is actually seen on screen, and there japanese arcade CRTs feature another white point calibration (D93 vs 'our' D65) and the gamma difference (arcade CRT vs sRGB) compared to our common displays.

dukeeeey:
If you clean up the code a bit I'll add it. Trying to clean up the whole code base and fix some of the more egregious code quality issues. For example we have uninitialized class variables all over the place and we've absolutely had bugs related to these in the past. For example older builds the audio was completely destroyed if you compiled with visual studio. The problem wasn't visual studio but the code itself. In the end I replaced the entire mpeg audio decoding lib.

But code readability and maintainability matters especially when people that made the pull requests are long gone.

toxieainc:
Please check if its now fine, thanks!

PR #197, opened 2024-09-25
Implement different upscaling filters for the 2D layers

toxieainc:
none, biquintic, bilinear (=old one), bicubic

All have their pros and cons, i personally favor either complete pixel sharpness (=none) or completely blurring away pixel borders (=bicubic), depending on the game

### DIFF
--- a/Config/Supermodel.ini
+++ b/Config/Supermodel.ini
@@ -42,6 +42,9 @@ Supersampling = 1
 ; 2=PVM_20M2U/D93, 3=BT601_525/D93, 4=BT601_525/D65 (recommended for all US developed games)
 ; 5=BT601_625/D65 (recommended for all EUR/AUS developed games)
 CRTcolors = 0
+; Upscale filter used for the 2D layers: 0=none/sharp pixels, 1=biquintic,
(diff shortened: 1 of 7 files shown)


dukeeeey:
Should bilinear be default? Honestly I've never even heard of biquintic. I know that bicubic can offer slightly better image quality but requires more texture reads so is slightly more expensive.

toxieainc:
Biquintic is based on a simple idea, popularized by iq: https://iquilezles.org/articles/texture/
It features smoother pixel 'borders', but looks overall a bit sharper than bilinear. I think it fits a bit better than bilinear. Will make some comparison shots.

toxieainc:
nearest
biquintic
bilinear
bicubic

toxieainc:
So, as you can see, the filters are basically increasing in 'perceived bluriness'.

dukeeeey:
On my phone the shader toy demo works at 1.3 fps lol
I'd say use the fast and dumb (bilinear) as default and if people want to experiment with others they can.

toxieainc:
In practice it should be almost for free compared to bilinear (its just a bunch of simple math ops), but i'll change it.

Issue #178, opened 2024-08-05
Driver timeout on AMD 5700U


toxieainc:
Hi!

Using Windows 11, an AMD 5700U (=integrated graphics) and latest official AMD drivers, Supermodel runs into a driver timeout pretty consistently, e.g. when selecting the race car in Daytona2PE (or during race in Scud Race).

This seems to be the case since at least the super sampling was introduced (note: happens also though if using supersampling = 1), but could also be some commits earlier on, i still need to do a full binary search.

dukeeeey:
Hi toxieainc,
if you could figure out at which commit this started happening that would help a lot. Unfortunately I don't even have an AMD card to test with. I know in windows there is a driver timeout value that you can edit in the registry. If you increase this time? 

https://answers.microsoft.com/en-us/windows/forum/all/increase-time-out-limit/e979e2ad-e15f-450b-9818-a148cbf01078

does it allow the game to run at least? Maybe something in the shader is is causing it to recompile eating up time.

If you are savvy with the code there try uncommenting these

  //glDebugMessageCallback(DebugCallback, NULL);
  //glDebugMessageControl(GL_DONT_CARE,GL_DONT_CARE,GL_DONT_CARE, 0, 0, GL_TRUE);
  //glEnable(GL_DEBUG_OUTPUT);

it should give some detailed driver output

toxieainc:
From my understanding, a recompile should be handled differently in the OS than the GPU not responding (as the latter prevents the screen to update, etc, while the first 'just' blocks some CPU thread(s)).

Will debug/find the commit when i'm back at that system (will take some weeks though :/).

dukeeeey:
Well opengl commands are essentially issued on a single thread (the current context). If swapbuffers takes too long, I think it's usually a second or two, windows assumes it's died and kills it. I know this can happen if you render very large datasets and it simply takes too long. But really it shouldn't happen in normal rendering. Uniforms are essentially constants per draw call, I know some vendors will optimise the shaders and essentially recompile them based upon different inputs. I dont know if this is happening here it's just speculation.

Really the best option is to enable those debug options and the driver will hopefully tell us what the issue is.

dukeeeey:
quad rendering?

toxieainc:
No, these changes/commits all worked fine, and also some months after that.

As said, will know more when i have access to the system again later-on.

crashGG:
[...]

PR #198, opened 2024-09-29
improve SuperSleepUntil implementation

toxieainc:
1) as found in lots of experiments done for the VPX and PinMAME projects, Sleep() on windows can oversleep for over 1ms, especially when doing Sleep(>1)
2) thus loop Sleep(1) and end if its less than 2ms
3) in the spin to wait for the rest of the time, insert yield(=_mm_pause) or the determined by the Rust devs arm64 equivalent

This actually gets rid of micro-stutter on my AMD based minipc e.g. in Daytona2 (see #178)

Then also use same Sleep() implementation in the network code (to avoid potential sideeffects between the 2 implementations)

### DIFF
--- a/Src/Network/SimNetBoard.cpp
+++ b/Src/Network/SimNetBoard.cpp
@@ -20,10 +20,10 @@
  ** with Supermodel.  If not, see <http://www.gnu.org/licenses/>.
  **/
 
-#include <chrono>
 #include <thread>
 #include "Supermodel.h"
 #include "SimNetBoard.h"
+#include <OSD/Thread.h>
 
  // these make 16-bit read/writes much neater
 #define RAM16 *(uint16_t*)&RAM
@@ -527,8 +527,6 @@ void CSimNetBoard::GetGame(const Game& gameInfo)
 
 void CSimNetBoard::ConnectProc(void)
 {
-	using namespace std::chrono_literals;
-
 	if (m_connected)
 		return;
 
@@ -546,7 +544,7 @@ void CSimNetBoard::ConnectProc(void)
 	{
 		if (m_quit)
 			return;
-		std::this_thread::sleep_for(1ms);
+		CThread::Sleep(1);
 	}
 
 	printf("Successfully connected.\n");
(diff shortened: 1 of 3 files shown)


toxieainc:
See our battle-tested implementation in PinMAME: https://github.com/vpinball/pinmame/blob/master/src/windows/ticker.c#L253

dukeeeey:
I'd much prefer a solution based upon SDL if possible, even if it's slightly sub optimal for different platforms. The original code does assume that windows sleep isn't that accurate, that's why it subtracts 1ms from the sleep time, then just spins for the rest of the time. But it's possible I guess that's not quite enough, and it goes over the time slightly.

The network code really doesn't need any kind of spin lock at all, std::this_thread::sleep_for(16ms); really is enough. The code could be changed to sleep for 1 second and it would still be fine.

toxieainc:
It is still all SDL based, also the network code (CThread::Sleep maps to SDL_Delay, so no spin there).
Its just the additional pause of the processor in the spin loop that is platform dependent that helps with hyper-threading, etc. To my knowledge there is no real equivalent in SDL or std.

toxieainc:
SDL3 will add one though apparently: https://wiki.libsdl.org/SDL3/SDL_CPUPauseInstruction

toxieainc:
Turns out that this was also added later-on for SDL2, just not documented.
So the code is now all SDL, so the PR now is just avoiding the potential additional 1ms too long Sleep (some outliers feature even more, if Sleep() is used with >1, thus the Sleep(1) loop) and the recommended CPU pause in the spin loop.
(and replacing the chrono sleep with the SDL Sleep/Delay).

toxieainc:
And as said, it fixes a real life issue on my mini PC, where with this changes some VERY noticable micro stutter vanishes.

dukeeeey:
SDL_CPUPauseInstruction is a nice solution. I think I need to update my sdl though because it's missing my from my version.

sgpowers:
What SDL2 version does this function "un-documentedly" appear in?

This will potentially break the the ability for some distributions to upgrade supermodel, particular in Linux.

toxieainc:
@sgpowers I tweaked it now, so that this part is optional. So this part will function like it did before, if SDL2 is too old.

PR #202, opened 2024-10-02
Static Analysis

toxieainc:
Addresses mostly harmless things found by static analysis while trying to nail down #178


Src/GameLoader.cpp
```diff
@@ -587,7 +587,7 @@ void GameLoader::IdentifyGamesInZipArchive(
       Region::ptr_t region = v2.second;
       if (!region->required)
         continue;
-      for (auto file: region->files)
+      for (const auto& file: region->files)
```
trzy, line 590: Can you make this "const auto &file". I prefer that style (and it's consistent with elsewhere in the code that I've written).

Src/GameLoader.cpp
```diff
@@ -757,7 +757,7 @@ bool GameLoader::ComputeRegionSize(uint32_t *region_size, const GameLoader::Regi
   // use maximum end_addr = offset + stride * (num_chunks - 1) + chunk_size.
   std::vector<uint32_t> end_addr;
   bool error = false;
-  for (auto file: region->files)
+  for (const auto& file: region->files)
```
trzy, line 760: Here too.

Src/Graphics/Legacy3D/TextureRefs.cpp
```diff
@@ -261,8 +261,7 @@ void CTextureRefs::DeleteAllHashEntries()
 	for (unsigned i = 0; i < m_hashCapacity; i++)
 	{
 		HashEntry *entry = m_hashEntries[i];
-		if (entry)
```
trzy, line 264: Are you sure this check isn't needed here?

Src/Graphics/Render2D.cpp
```diff
@@ -411,7 +411,7 @@ void CRender2D::AttachVRAM(const uint8_t* vramPtr)
 {
 }
 
-void CRender2D::AttachDrawBuffers(std::shared_ptr<TileGenBuffer> bottom, std::shared_ptr<TileGenBuffer> top)
+void CRender2D::AttachDrawBuffers(const std::shared_ptr<TileGenBuffer>& bottom, const std::shared_ptr<TileGenBuffer>& top)
```
trzy, line 414: Nit: Move & to align with left of parameter name.

Src/Graphics/Render2D.h
```diff
@@ -148,9 +148,9 @@ class CRender2D
 	*
 	* Parameters:
 	*    bottom		Bottom surface
-	*	 top		Top surface	
+	*    top		Top surface	
 	*/
-	void AttachDrawBuffers(std::shared_ptr<TileGenBuffer> bottom, std::shared_ptr<TileGenBuffer> top);
+	void AttachDrawBuffers(const std::shared_ptr<TileGenBuffer>& bottom, const std::shared_ptr<TileGenBuffer>& top);
```
trzy, line 153: Nit: &bottom, &top

Src/Inputs/Inputs.cpp
```diff
@@ -316,9 +316,9 @@ CGearShift4Input *CInputs::AddGearShift4Input(const char *id, const char *label,
 }
 
 CTriggerInput *CInputs::AddTriggerInput(const char *id, const char *label, unsigned gameFlags, 
-	CSwitchInput *trigger, CSwitchInput *offscreen, UINT16 offVal, UINT16 onVal)
+	CSwitchInput *_trigger, CSwitchInput *offscreen, UINT16 offVal, UINT16 onVal)
```
trzy, line 319: Why this?

Src/Model3/53C810.h
```diff
@@ -194,19 +194,19 @@ class C53C810: public IPCIDevice
...
 	void	Insert(UINT8 mask, UINT8 op, Result (*Handler)(struct NCR53C810Context *));
 	Result	(*OpTable[256])(struct NCR53C810Context *);
-	
+
 	// Context (register file)
 	struct NCR53C810Context	Ctx;
 	
 	// IRQ controller and IRQ identifier for this SCSI controller
-	CIRQ		*IRQ;
+	//CIRQ		*IRQ;
```
trzy, line 209: If CIRQ isn't needed, better to just delete it than have it commented out.

Src/Network/NetBoard.h
```diff
@@ -99,7 +99,7 @@ class CNetBoard : public IBus, public INetBoard
 	UINT16		recv_size;
 	UINT16		send_offset;
 	UINT16		send_size;
-	UINT8		slot;
```
trzy, line 102: If truly unused, delete.

Src/Network/NetBoard.h
```diff
@@ -114,7 +114,7 @@ class CNetBoard : public IBus, public INetBoard
 
 	// only for some tests
 	UINT8 *bank;
-	UINT8 *bank2;
+	//UINT8 *bank2;
```
trzy, line 117: If unused, delete.

Src/OSD/Logger.cpp
```diff
@@ -295,7 +295,7 @@ void CFileLogger::ReopenFiles(std::ios_base::openmode mode)
   m_logFiles.clear();
 
   // (Re-)Open
-  for (auto filename: m_logFilenames)
+  for (const auto& filename: m_logFilenames)
```
trzy, line 298: const auto &filename

Src/OSD/Logger.cpp
```diff
@@ -318,14 +318,14 @@ void CFileLogger::WriteToFiles(const char *str)
   }
 }
 
-CFileLogger::CFileLogger(CLogger::LogLevel level, std::vector<std::string> filenames)
+CFileLogger::CFileLogger(CLogger::LogLevel level, const std::vector<std::string>& filenames)
```
trzy, line 321: &filenames

Src/OSD/Logger.cpp
```diff
@@ -318,14 +318,14 @@ void CFileLogger::WriteToFiles(const char *str)
...
-CFileLogger::CFileLogger(CLogger::LogLevel level, std::vector<std::string> filenames)
+CFileLogger::CFileLogger(CLogger::LogLevel level, const std::vector<std::string>& filenames)
   : m_logLevel(level),
     m_logFilenames(filenames)
 {
   ReopenFiles(std::ios::out);
 }
 
-CFileLogger::CFileLogger(CLogger::LogLevel level, std::vector<std::string> filenames, std::vector<FILE *> systemFiles)
+CFileLogger::CFileLogger(CLogger::LogLevel level, const std::vector<std::string>& filenames, const std::vector<FILE *>& systemFiles)
```
trzy, line 328: & alignment

Src/OSD/Logger.h
```diff
@@ -183,13 +183,13 @@ class CFileLogger: public CLogger
   void DebugLog(const char *fmt, va_list vl);
   void InfoLog(const char *fmt, va_list vl);
   void ErrorLog(const char *fmt, va_list vl);
-	CFileLogger(LogLevel level, std::vector<std::string> filenames);
-  CFileLogger(LogLevel level, std::vector<std::string> filenames, std::vector<FILE *> systemFiles);
+  CFileLogger(LogLevel level, const std::vector<std::string>& filenames);
```
trzy, line 186: & alignment

Src/OSD/Logger.h
```diff
@@ -183,13 +183,13 @@ class CFileLogger: public CLogger
   void DebugLog(const char *fmt, va_list vl);
   void InfoLog(const char *fmt, va_list vl);
   void ErrorLog(const char *fmt, va_list vl);
-	CFileLogger(LogLevel level, std::vector<std::string> filenames);
-  CFileLogger(LogLevel level, std::vector<std::string> filenames, std::vector<FILE *> systemFiles);
+  CFileLogger(LogLevel level, const std::vector<std::string>& filenames);
+  CFileLogger(LogLevel level, const std::vector<std::string>& filenames, const std::vector<FILE *>& systemFiles);
```
trzy, line 187: & alignment

Src/Sound/SCSPDSP.cpp
```diff
@@ -479,13 +479,14 @@ void SCSPDSP_Start(_SCSPDSP *DSP)
 	}
 */
 
-	for(int t=0;t<0x10000;++t)
+/*	for(int t=0;t<0x10000;++t)
```
trzy, line 482: Why is this loop being commented out?

Src/Util/GenericValue.h
```diff
@@ -29,19 +29,19 @@ namespace Util
     template <typename T>
     struct IntegerEncodableAsHex
     {
-      static const bool value = std::is_integral<T>::value && sizeof(T) >= 2 && sizeof(T) <= 8;
```
trzy, line 32: It's been a long time since I've looked at this template stuff but I thought these ::value and ::type members were required?

Src/Util/NewConfig.cpp
```diff
@@ -238,7 +238,7 @@ namespace Util
       if (Exists())
         m_value->Serialize(os);
       *os << "\"  children={";
-      for (auto v: m_children)
+      for (const auto& v: m_children)
```
trzy, line 241: const auto &v:

trzy (commented):
Mostly nitpicks, a few questions about code being deleted.

Src/Graphics/Legacy3D/TextureRefs.cpp
toxieainc: Yes, nowadays all compilers respect the c++ standard, that says 'In all cases, if ptr is a null pointer, the standard library deallocation functions do nothing.'

Src/Graphics/Render2D.cpp
toxieainc: I guess @dukeeeey is right, this is maybe a bit stupid

Src/Sound/SCSPDSP.cpp
toxieainc: It is simply debug code it seems, as it has no effect in practice (only if one would set a breakpoint in there or the like).

Src/Util/GenericValue.h
toxieainc: The analyzer recommended to modernize it, so it should be equivalent.

toxieainc:
All should be addressed, thanks!

Src/Inputs/Inputs.cpp
toxieainc: To avoid confusion, as the class already contains a 'trigger'

toxieainc:
@trzy Is this okay like this?

trzy:
Looks good. Thank you!

toxieainc:
Another question regarding constructors, do you prefer the 'classic' initialization a la
```
Bla::Bla() : m_member1(0), m_member2(nullptr), ... ...
{
   memset(...);
   ...
}
```
or via default member initialization directly in the class/member declaration
```
class Bla
{
   uint m_member1 = 0;
   void* m_member2 = nullptr;
   ...
}
```
Nowadays i personally prefer the latter as then its easier to spot if something has been forgotten, and also the order is clear, if there are dependencies.

dukeeeey:
I normally use the latter for structs especially if they have no constructor at all, and the former for classes. If all elements are set to zero or null it just kind of needlessly fattens the class header. But this is just personal preference, I don't think there is a right and wrong answer for this.

toxieainc:
Okay, as i'm currently preparing a PR for more class inits, what to do in case if there is currently a mixture? So some members use a), some b).

dukeeeey:
If it has a mixture I mean you can refactor if you want but there isn't a strong need for it. As long as the classes are initialised to a known state. 

Some of the structs used for rendering its preferable for performance that they not be initialised. All members will be written to anyway.

toxieainc:
Absolutely, the 'small' critical ones i will not touch..

Issue #203, opened 2024-10-03
Visual studio build quirk(s)


toxieainc:
Using my own VS2022/2019 builds, e.g. both Daytona 2 games and Scud Race feature a quirk where it seems the gas pedal (all pedals?) is stuck pressed at the beginning of the race, but only on certain tracks (e.g. on D2 medium/advanced course, SR expert course).

This does not happen with the official gcc based build.

dukeeeey:
My guess is our friend the uninitialized variable. I fixed a whole bunch but there are more.

gm-matthew:
Maybe it's worth trying a clean build? Try compiling with Rebuild Solution instead of Build Solution and see if it changes anything.

Occasionally I'll get weird issues - usually Supermodel failing to start - that are fixed by making a clean build. Also I personally turn off link-time code generation (LTCG) since that seems to make corrupt builds less frequent.

toxieainc:
Yes, i already tried all kinds of things. Also changing various compiler options, and trying to init more class members (just to be sure), but so far the problem remains with the VS builds.

toxieainc:
I guess next step would be to run a real mem tracker like Dr.Memory on it to see where/if uninit'ed mem is used or if there is some other unknown glitch.

toxieainc:
So far no luck with uninit'ed mem.  :/
The tool reports tons of issues in the Intel GFX driver though  ;)

trzy:
Have you tried some good old fashioned printf() debugging to see where the input values appear to be corrupted? Starting to suspect a code generation issue and unfortunately, it's somewhere in the CPU emulator, it'll be hard to find, but you'll be able to assess at least whether the input system is returning the expected values (which would indicate something is happening downstream in the CPU emulator). I haven't tried building with MSVC yet but this shouldn't be too hard to debug. I may be able to look into it after Monday. Does it also happen when the debugger is built? If so, the debugger will be immensely helpful. Input values are usually read once per frame in a single place.

toxieainc:
@trzy ReadInputs() does not show any activity for Steer,Acc or Break for the m_game.inputs & Game::INPUT_VEHICLE case.
Or is there more to check?

trzy:
This sounds bad. But: does it show activity when compiled with gcc or in the stages where it works? If so (and I do think it should unless I am missing something obvious), then I think the problem is inside the input system.

toxieainc:
[...]

PR #205, opened 2024-10-05
More static analysis fixes

toxieainc:
and one more in the hope of fixing #178 or #203, unfortunately no luck with those


dukeeeey:
const unsigned* const __restrict 

That's not standard c++.

toxieainc:
True, but pretty much all common c++ compilers accept it, we also use it in all commercial code i work on for a living (which covers pretty much all x86/arm/ppc platforms in use nowadays). In fact, most c++ compilers generate less efficient code without it.
But if you prefer, i can also remove it, of course.

dukeeeey:
Supermodel ends up on all kinds of platforms you wouldn't expect, xbox etc. Would prefer to use standard functions. But the texture loading code is not a hot path. In such cases readability is more preferable.

toxieainc:
Removed again..

Issue #206, opened 2024-10-05
Scaling issue


toxieainc:
Starting with 905bfa2 (8e7b850 still worked as before), it seems like the scaling of the framebuffer is handled differently, although there is no explanation of why this happened suddenly (i didn't see anything fishy in the commits), e.g. when running
supermodel -res=1920,1080 srally2.zip

Another uninitialized var problem?

dukeeeey:
hm I can't replicate this.
What settings? AA? Does it happen in windowed mode? GPU ?

toxieainc:
Just the raw command line above, default ini file, so windowed mode.

dukeeeey:
What windows dpi

toxieainc:
No clue, 125% scaling.

toxieainc:
But that is the same in both tests, so should not matter, right? Or was some build config changed in that range?

dukeeeey:
I *think* dpi can change the size of the window you render to. So maybe this is the issue. Try 100% scaling

toxieainc:
Will try, still confusing if that would be the cause though, unless the build setup changed/tools were updated, otherwise all should react the same, right?

dukeeeey:
![image](https://github.com/user-attachments/assets/55c11407-67b6-45a3-a8be-26ae37705938)
what mine looks like?

toxieainc:
and without wide screen?

toxieainc:
https://github.com/trzy/Supermodel/commit/8e7b8506ff0c0f2b0ed994205e4108bc387fab16: ![grafik](https://github.com/user-attachments/assets/6deea7f6-5219-4490-b11a-8f51abd45ed9)

toxieainc:
Okay, i nailed it down now. As i have a 2GPU laptop setup (Intel CPU/GPU + NV GPU), for whatever reason the NV does the weird scale, the Intel the expected one.  :/
Why windows decided to pick 1.GPU for that build, and 2.GPU for the other (same executable name) is beyond me.

dukeeeey:
Could be some driver option? Honestly I'm not sure

toxieainc:
Got it. Its really related to the windows DPI. If one embeds in the executable that the app is DPI aware then it works.
Will do a PR to change the vcproj, and will try to find out how that works for gcc.

toxieainc:
Apparently NV driver does the DPI scale, Intel not (don't know which is correct, as this is a GL windows after all).

PR #208, opened 2024-10-08
DPI awareness

toxieainc:
Flag SDL window as DPI-aware, otherwise the window content might be scaled by some windows graphics drivers

addresses #206


dukeeeey:
Nice fix :)

Issue #201, opened 2024-10-02
Sega rally desert stage bug


Garagefreek:
The beginning of the race there’s a vertical line at the centre of the screen then disappears when you start driving?

Garagefreek:
![Screenshot 2024-10-02 010537](https://github.com/user-attachments/assets/bf374b98-18eb-4b64-b08a-d7fc3b095d82)
As far as i know this only happens on a 1920x1080 display only ??

jaoxford:
Duplicate to this issue: https://github.com/trzy/Supermodel/issues/199

toxieainc:
Which build did you use for testing?

jaoxford:
Try changing resolutions, see if it happens at one in particular.

Does this bug happen on only the latest Supermodel build, or have you tried an older build?

Garagefreek:
It's happening on the latest build and the previous one on the 1920x1080 display only, it works fine on every other displays.

gm-matthew:
As a temporary fix you could try setting the resolution to 1920x1079, that should get rid of the vertical line

Garagefreek:
Will do, thank you.

toxieainc:
Seems like this issue has been around for quite some time at least

gm-matthew:
I've discovered that this bug only occurs when using the quad renderer; with the triangle renderer it looks fine.

The vertical line is a gap in the skybox model which shouldn't be there as there are no gaps between polygons. There must be some quirk of the quad renderer that is drawing the polygons a tiny bit smaller than they are supposed to be.

toxieainc:
Thanks, that helps. I'm currently looking at the code again and already saw some spots where things could be simplified, hopefully improving precision.

dukeeeey:
The line at the top I'm sure is a precision issue with the quad renderer. The attribute interpolation on the edge is bad. I had problems like this with my original code and my fix was to use double precision maths to calculate the area of the poly. Of course this doesn't play well with older hw.

toxieainc:
Maybe i can tweak it once more, maybe also helping the old Harley problem, too.

toxieainc:
@dukeeeey Can you please point me to games/scenes where the 
```
if (lambdaSignCount != 4) {
	if(!gl_HelperInvocation) {
		discard;
	}
}
```
can be seen in action? Cause i struggle at the moment to find anything where simply commenting out that discard does change anything.

dukeeeey:
[...]

PR #212, opened 2024-10-13
Improve precision for vp computation

toxieainc:
...to fix regression for at least SRally2 and HarleyD

fixes/improves #201

### DIFF
--- a/Src/Graphics/New3D/Mat4.cpp
+++ b/Src/Graphics/New3D/Mat4.cpp
@@ -151,12 +151,12 @@ void Mat4::Rotate(float angle, float x, float y, float z)
 
 void Mat4::Frustum(float left, float right, float bottom, float top, float nearVal, float farVal)
 {
-	float x = 2.0F / (right - left);
-	float y = 2.0F / (top - bottom);
-	float a = (right + left) / (right - left);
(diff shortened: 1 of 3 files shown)


dukeeeey:
Are these functionally the same?

	float x = 2.0F / (right - left);
	float y = 2.0F / (top - bottom);
	float a = (right + left) / (right - left);
	float b = (top + bottom) / (top - bottom);

	float x = 2.0f / (right - left);
	float y = 2.0f / (top - bottom);
	float a = left*x + 1.0f;
	float b = bottom*y + 1.0f;

Some games use off-axis projection, harley etc

toxieainc:
I personally prefer it like that, so that its clear whats happening. And in theory you can loose precision, depending on the integer value.

And the code is the same, yes. This way you can share more of the same potential error.

toxieainc:
As for putting additional stuff into such PRs: Sorry, i will try to be more careful then.
I have a lot of local changes piled up, so if some are in the same file as my 'actual fix', and these are rather harmless, then i tend to just squeeze them in, too.

dukeeeey:
I've pushed your changes here locally :) Without the changes for the matrix stuff. I'll test the other code soon.

PR #213, opened 2024-10-13
Improve edge test precision of quad test, and add missing '----'-area

toxieainc:
...support for twisted/bow-tie quads

..and align triangle code more to the quad code for easier comparisons

also generalize mipmap0-only optimization to all non-blended cases

related to #201

### DIFF
--- a/Src/Graphics/New3D/R3DShaderCommon.h
+++ b/Src/Graphics/New3D/R3DShaderCommon.h
@@ -14,32 +14,32 @@ vec4 ExtractColour(int type, uint value)
 {
 	vec4 c = vec4(0.0);
 
-	if(type==0) {			// T1RGB5	
-		c.r		= float((value >> 10) & 0x1Fu);
-		c.g		= float((value >> 5 ) & 0x1Fu);
-		c.b		= float((value      ) & 0x1Fu);
+	if(type==0) {			// T1RGB5
+		c.r     = float((value >> 10) & 0x1Fu);
+		c.g     = float((value >> 5 ) & 0x1Fu);
+		c.b     = float((value      ) & 0x1Fu);
 		c.rgb  *= (1.0/31.0);
-		c.a		= 1.0 - float((value >> 15) & 0x1u);
+		c.a     = 1.0 - float((value >> 15) & 0x1u);
 	}
 	else if(type==1) {		// Interleaved A4L4 (low byte)
-		c.rgb	= vec3(float(value&0xFu));
-		c.a		= float((value >> 4) & 0xFu);
+		c.rgb   = vec3(float(value&0xFu));
+		c.a     = float((value >> 4) & 0xFu);
 		c      *= (1.0/15.0);
 	}
 	else if(type==2) {
-		c.a		= float(value&0xFu);
+		c.a     = float(value&0xFu);
 		c.rgb   = vec3(float((value >> 4) & 0xFu));
-		c	   *= (1.0/15.0);
+		c      *= (1.0/15.0);
 	}
 	else if(type==3) {
-		c.rgb	= vec3(float((value>>8)&0xFu));
-		c.a		= float((value >> 12) & 0xFu);
-		c	   *= (1.0/15.0);
+		c.rgb   = vec3(float((value>>8)&0xFu));
+		c.a     = float((value >> 12) & 0xFu);
+		c      *= (1.0/15.0);
 	}
 	else if(type==4) {
-		c.a		= float((value>>8)&0xFu);
+		c.a     = float((value>>8)&0xFu);
 		c.rgb   = vec3(float((value >> 12) & 0xFu));
-		c	   *= (1.0/15.0);
+		c      *= (1.0/15.0);
 	}
 	else if(type==5) {
 		c = vec4(float(value&0xFFu) / 255.0);
@@ -145,7 +145,6 @@ float LinearTexLocations(int wrapMode, float size, float u, out float u0, out fl
 		return fract(u);						// return weight
 	}
 	else {										// mirror + mirror clamp - both are the same since the edge pixels are repeated anyway
-
 		float odd = floor(mod(u, 2.0));			// odd values are mirrored
 
 		if(odd > 0.0) {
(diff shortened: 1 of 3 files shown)


dukeeeey:
Hey, sorry super long time come back to this. My real life basically hasn't allowed me to work on any side projects for some time. (Had a health disaster). I had some thoughts about this, and did some debugging. I think the basic problem is the maths just falls over on in these edge cases, not because the algorithm is necessarily bad but due to I think limited precision. You fixed the area precision part, or at least it works pretty well with floating point. Anyway from what I can see the problem only effects the edge pixels. When rendering, the interpolation value say for a triangle, the value is always a mixture of the attributes from each of the 3 vertices. This is true for every pixel in a triangle.. except the edge pixels. Edge pixels are sort of a special case because you essentially are only interpolating between 2 vertices, the barycentric coordinate for the 3rd should be zero so the interpolation from the 3rd vertex is essentially zero. The same is true for quads. I wonder if we could leverage this and just replace the 4 point interpolation with essentially 2 point interpolation for the edge pixels. There is also no culling require I don't think for edge pixels.

toxieainc:
Oh, i hope you are in good shape again and feel better by now!  :/

The idea with interpolating only along the edge sounds nice, but i wonder how to detect this case robustly. But maybe just checking for something close to 0 may really could do the trick.

dukeeeey:
I think basically to do this we would pass barycentric coordinates. Basically pass an extra vec3 for each vertex in the triangle. So something like vertex1 = vec3(1,0,0) vertex 2 = vec3(0,1,0), vertex 3 = vec3(0,0,1)

I did the maths and I think supermodel is using 15 vertex slots, there is a usual max of 16 for gl 3.3 class h/w. So we have 1 slot free lol.

Basically to detect the edge pixels one of the coordinates will be 0. I might need to write some test code for this, but I think it should work.

dukeeeey:
Should add, we could pass negative coordinates for the 2nd triangle, to differentiate the two triangles that make up the quad. So we we know which vertices to look up.

toxieainc:
From quickly diving into this again, i think we have 2 different issues here:
a) precision for drawing the 'stand-in' geometry (so most likely VS or GS part)
b) precision when discarding pixels (FS)

So both a) and b) seem to cause parts of the issues one sees in srally2.

dukeeeey:
Have a look at these. 
First pic I passed the barycentric triangle coordinates, then wrote them as RGB
image

2nd pic. Just basic detection for quad edges using these coordinates. I set the quad edge colours to the barycentric coordinates

	if(fs_in.barycentricCoords.x<0.01 || fs_in.barycentricCoords.z<0.01) {
		fsColor = vec4(fs_in.barycentricCoords,1.0);
	}

image

Last pic, I tweaked the condition to just draw with extreme values

	if(fs_in.barycentricCoords.x<0.0000001 || fs_in.barycentricCoords.z<0.0000001) {
		fsColor = vec4(fs_in.barycentricCoords,1.0);
	}

image
It literally just only effects the very center pixels in a vertical line, the exact same place we have this interpolation problem

dukeeeey:
So yeah my suggestion for this is to just do 2 point interpolation on the extreme edges
We can just plug in the barycentric coordinates directly I think as a look up

toxieainc:
Interesting (and nice results)! Maybe i screwed something up in my own experiments. Will you implement/finalize this then?

toxieainc:
From my understanding of how you want to tackle it, this should also be able to handle bowtie quads, etc, right?

dukeeeey:
Well I had a go at this. Basically it works.. but the problem is if i add any conditional logic with regards to writing to the interpolated vertex shader values it just seems to blow up the shader. So maybe back to the drawing board

If you look here this is a discard so these pixels would never even draw anyway

image

If i uncomment writing to fsViewVertex the shader goes crazy. Even though this path shouldn't be getting called.

image

toxieainc:
Do you have a branch of this somewhere? I could also have a look if you want.

dukeeeey:
Code is here if u want to give it a spin:
https://pastebin.com/gJc9R8YN

The points to look at are

    vec3 barycentricCoords;

Added to the geometry shader

Then the interpolation code in the fragment shader which looks like this

`        for (int i=0; i<3; i++) {
 
            int index       = indices[i];
            float w         = 1.0 / fs_in.oneOverW[index];
            float baryCoord = abs(fs_in.barycentricCoords[i]);
                
            fsViewVertex    += baryCoord * (fs_in.viewVertex[index] * w);
            fsViewNormal    += baryCoord * (fs_in.viewNormal[index] * w);
            fsTexCoord      += baryCoord * (fs_in.texCoord[index] * w);
            fsFixedShade    += baryCoord * (fs_in.fixedShade[index] * w);
            //interp_oneOverW   += baryCoord * (fs_in.oneOverW[index]);      
        }`

I think I'm either bumping into the limit for max vertex attribs, or the shader is too complex somehow for the h/w.

I left the 4 point interpolation code commented out, with just the 3 point interpolation doing it's thing. My idea was to just fallback to this code at the extreme edges of the triangles, this is where the quad code seems to fail completely.

toxieainc:
Thanks! Will have a look later-on if i can spot something.

dukeeeey:
Great, if we can't get this to work any sense I actually have a different idea for rendering, a sort of hybrid software/hardware style. Basically to software rasterise the poly edges then draw horizontal spans with pairs of thin triangles. We can emulate much closer to how the hardware actually draws this way, but will be more expensive CPU and probably GPU side.

The down side for the hardware in passing scan lines size triangles is hardware is really built to render blocks of 2x2 pixels so thin triangles are sub optimal.

toxieainc:
Oh, and, on which HW do you see this issue?

dukeeeey:
My dev machine has a nvidia rtx 3070

dukeeeey:
I am not sure if it's some sort of driver bug
Or some limitation that I am not aware of

dukeeeey:
Amazing work my friend. Only thing I would change is to const float fMin = 0.00006103515625; and use that instead of magic numbers everywhere. But otherwise great job. I did try and rework the code myself but it still wasn't playing ball.

toxieainc:
Should i do a PR then, or do you just want to commit it right away?

toxieainc:
Great, thanks!!

PR #215, opened 2024-10-20
properly enable fesetround/fegetround usage

### LINKED ISSUE #203: Visual studio build quirk(s)

toxieainc:
..according to the C++ spec

unfortunately, e.g. MSVC ignores this pragma, so use what MSDN suggests in that case (GCC and clang may still require -frounding-math for compilation of at least this file)

in addition, do not set fesetround mode once, but just each time before each math op that requires it otherwise other Supermodel code may use the wrong rounding modes (as it relies by default on it being FE_TONEAREST)

this fixes #203, maybe other emulation+math precision issues, too

### DIFF
--- a/Src/CPU/PowerPC/ppc_ops.c
+++ b/Src/CPU/PowerPC/ppc_ops.c
@@ -2184,7 +2184,7 @@ static void ppc_fctiwx(UINT32 op)
 		// FPSCR[FI] = 1
 		// FPSCR[XX] = 1
 	}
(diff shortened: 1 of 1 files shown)


toxieainc:
will redo the PR to fix the wrong commit message, sorry   :/

PR #216, opened 2024-10-21
Properly enable fesetround/fegetround usage

### LINKED ISSUE #203: Visual studio build quirk(s)

toxieainc:
..according to the C++ spec

unfortunately, e.g. MSVC ignores this pragma, so use what MSDN suggests in that case (GCC and clang may still require -frounding-math for compilation of at least this file)

in addition, do not set fesetround mode once, but just each time before each math op that requires it otherwise other Supermodel code may use the wrong rounding modes (as it relies by default on it being FE_TONEAREST) note that changing the rounding mode is extremely rare, so this will not affect Supermodels performance by much

and fix additional minor issue in ppc_fctiwx

this fixes #203, maybe other emulation+math precision issues, too

### DIFF
--- a/Src/CPU/PowerPC/ppc_ops.c
+++ b/Src/CPU/PowerPC/ppc_ops.c
@@ -40,8 +40,6 @@
 #pragma fenv_access(on) // because of fesetround
 #endif
 
-static unsigned int rounding_mode = FE_TONEAREST;
-
 static void ppc_unimplemented(UINT32 op)
 {
 	ErrorLog("PowerPC hit an unimplemented instruction. Halting emulation until reset.");
@@ -1614,36 +1612,28 @@ inline bool sign_double(FPR x)
 	return (x.id >> 63);
 }
 
-// in theory the following 3 functions require compiler options to work correctly
+// in theory the following 2 functions require compiler options to work correctly
 //  -frounding-math for GCC
 //   GCC and clang can show weird behavior nevertheless according to various threads on the net
 //  /fp:strict for MS Visual Studio/C++
 //   here we use a set of pragmas at the beginning of the file instead (to steer selective behavior for just this file)
 // 
 // unknown if any games actually change this, at least its extremely rare for Model3
-inline void set_rounding_mode(void)
-{
-	switch (ppc.fpscr & 3)
-	{
-	case 0: rounding_mode = FE_TONEAREST; break;
-	case 1: rounding_mode = FE_TOWARDZERO; break;
-	case 2: rounding_mode = FE_UPWARD; break;
-	case 3: rounding_mode = FE_DOWNWARD; break;
-	}
-}
 
 // we just 'cache' FE_TONEAREST, as other math functions that are called
 // by other code inbetween PPC emulation could be influenced in a bad way,
 // as these may expect FE_TONEAREST to be set
 inline void init_rounding_mode(void)
 {
-	if (rounding_mode != FE_TONEAREST)
-		fesetround(rounding_mode);
+	static constexpr unsigned int rounding_mode[4] = {FE_TONEAREST,FE_TOWARDZERO,FE_UPWARD,FE_DOWNWARD};
+
+	if ((ppc.fpscr & 3) != 0) // rounding mode not nearest?
+		fesetround(rounding_mode[ppc.fpscr & 3]);
 }
 
 inline void restore_rounding_mode(void)
 {
-	if (rounding_mode != FE_TONEAREST)
+	if ((ppc.fpscr & 3) != 0) // rounding mode not nearest?
 		fesetround(FE_TONEAREST);
 }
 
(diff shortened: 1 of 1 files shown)


toxieainc:
@trzy As you wanted to know if the MSVC issue with scud race and daytone has been pinpointed now:
Yes, this is the case. The actual problem was in `fselx`, as it did ignore NaNs. So that part is now explicitly done in case the compiler switches the comparison around or the like (which now also follows the fselx spec exactly).

Weird enough, the fixes for feset/fegetround usage did also fix this (which must be a sideeffect/compiler-heuristic-qwirk of MSVC) though. I still left this in though, as it now follows the recommendations from MSDN and the C++ spec when wanting to change rounding modes.

In addition, the PR avoids that fesetround may change the behavior of all other Supermodel floating point code, by changing it only when needed (which is extremely rare).

Then it also changes some bugs in qNaN detection, now also follows the PPC spec.

Plus fctiw(z)x had one minor and one medium bug when doing the clamp.

toxieainc:
@trzy @dukeeeey Is this okay as-is? All my testing so far also worked fine..

dukeeeey:
It looks good to me but powerpc code is not my strong point so I'll let Bart review and merge these changes :)

Src/CPU/PowerPC/ppc_ops.c
```diff
@@ -2720,7 +2752,7 @@ static void ppc_fselx(UINT32 op)
 
 	CHECK_FPU_AVAILABLE();
 
-	FPR(t).fd = (FPR(a).fd >= 0.0) ? FPR(c).fd : FPR(b).fd;
+	FPR(t).fd = (is_nan_double(FPR(a)) || FPR(a).fd < 0.0) ? FPR(b).fd : FPR(c).fd; // do not just sign check, as -0.0 also counts as +0.0
```
trzy, line 2755: I don't understand the check for NaN here being related to -0.0? If this is the actual behavior, that's fine, but is it?

Src/CPU/PowerPC/ppc_ops.c
```diff
@@ -1614,67 +1571,82 @@ inline INT64 round_toward_negative_infinity(FPR f)
...
 }
 
+// in theory the following 3 functions require compiler options to work correctly
+//  -frounding-math for GCC
+//   GCC and clang can show weird behavior nevertheless according to various threads on the net
+//  /fp:strict for MS Visual Studio/C++
+//   here we use a set of pragmas at the beginning of the file instead (to steer selective behavior for just this file)
+// 
+// unknown if any games actually change this, at least its extremely rare for Model3
 inline void set_rounding_mode(void)
```
trzy, line 1624: Why is set_rounding_mode() separate from init_rounding_mode()? Is the intermediate rounding_mode var needed? Why not check fpscr directly in init_rounding_mode()?

Src/CPU/PowerPC/ppc_ops.c
```diff
@@ -1614,67 +1571,82 @@ inline INT64 round_toward_negative_infinity(FPR f)
...
+inline void init_rounding_mode(void)
+{
+	if (rounding_mode != FE_TONEAREST)
+		fesetround(rounding_mode);
+}
+
+inline void restore_rounding_mode(void)
+{
+	if (rounding_mode != FE_TONEAREST)
+		fesetround(FE_TONEAREST);
```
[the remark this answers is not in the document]
toxieainc: The initial state from my understanding is undefined, but all common compilers apparently assume nearest, as its also the default mode for x86 and arm.

Src/CPU/PowerPC/ppc_ops.c
toxieainc: Its not related to the -0 check, the 'old' check assumed that the compiler will not change the comparison, so the NaN test was included there. So i made it explicit to avoid this (as the PPC doc states for fselx).

Src/CPU/PowerPC/ppc_ops.c
toxieainc: Good point, will do it like that.

toxieainc: Done

trzy (approved):
Approving and merging. I think it looks ok. I tested a few games -- not as exhaustively as I would have liked -- and no perceivable difference except that I did encounter a "meow" sound twice in Scud Race playing the Beginner Night track. That sound effect is from Scud Race Plus and this is clearly a bug but I can't prove whether it's related to this change. It happens randomly, so hard to judge against current master. Let's keep an eye out for any unusual behavior.

PR #217, opened 2024-10-30
Add xdg base dirs

jflatt:
This will preserve existing ~/.config/supermodel/ and ~/.local/share/supermodel/ directories, and allow use of XDG base directories if those aren't present

### DIFF
--- a/Src/OSD/FileSystemPath.h
+++ b/Src/OSD/FileSystemPath.h
@@ -21,7 +21,7 @@
 
 /*
  * FileSystemPaths.h
- * 
+ *
  * Header file for OS-dependent Supermodel files locations.
  */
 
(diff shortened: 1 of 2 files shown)


trzy:
I am still not fully sure I understand why this is needed. Why do we want Supermodel's files to be scattered to the four winds all over the hard drive?

It's way easier to:

```
vim Config/Supermodel.ini
```

Than 

```
vim ~/.supermodel/config/Supermodel.ini
```

Emulators don't strike me as the sort of application that should be installed globally outside of the home directory anyway. Is there really a use case for this?

jflatt:
I fully agree!  Unfortunately this comes up a lot in the flatpak groups.  I suppose the plus side is that all other configuration are stored within a centralized location.  A use case could be backing up all config in a single folder (I don't know?).  In this commit, the existing search order is preserved (home, current directory, ~/.config), this one is just added to the bottom of the list.

PR #221, opened 2024-12-30
Appimage

dafrenchyman:
# Overall Summary:

PR to create an appimage along with a Github action that generates a **latest** release with the `x86_64` based image attached. You can see it's successful creation here: https://github.com/dafrenchyman/Supermodel/releases/tag/latest

## NOTE:
- I may be copying way too many libraries over (or I may also be missing some) to the app-image. I don't know what libraries should be available system-wide and which ones should be packaged in.

# Commit Notes:
## 1.) Preparing for AppImage creation
- Couldn't get the code to compile, had to add `#ifdef` around "SDL_HINT_WINDOWS_DPI_AWARENESS" in Src/OSD/SDL/Main.cpp
- Added a new `APP_IMAGE` flag to `Makefiles/Options.inc`
- If flag is set, new options get set to `PLATFORM_LDFLAGS`

## 2.) AppImage addition
- Added the generated and downloaded files to the .gitignore so they won't accidentally get added
- `AppImage/build_app_image.sh`: Bash script that automates the process of building out the app-image. Much easier to script it than have instructions
- `AppImage/Supermodel.png`: Couldn't find an *official* Supermodel logo. Found a Sega Model 3 logo and resized it to AppImage specifications
  - Should be replaced with a *real* Supermodel logo in the future
- `AppImage/Supermodel.desktop`: File required by the app-image
- `AppImage/AppRun`: Bash script that is actually run by the app-image

## 3.) Github action to build the app-image
- Ran into some issues with running the appimage directly via github.
- Runs correctly if it's contents are extracted and ran that way instead


Src/OSD/SDL/Main.cpp
```diff
@@ -2059,7 +2059,9 @@ int main(int argc, char **argv)
   }
 
   // Flag as DPI-aware, otherwise the window content might be scaled by some graphics drivers
+#ifdef SDL_HINT_WINDOWS_DPI_AWARENESS
```
dafrenchyman, line 2062: I had to add this or it wouldn't compile on linux

dafrenchyman:
Anyone around?

trzy:
I don't have access to a Linux machine to test this. I'll be able to take a look next time I go back to Nevada ca. October or November. But it looks like this needs updating. 

Can someone explain what this is, why it is needed, and why the existing Linux builds are insufficient? The Supermodel build process is really quite simple and I'm surprised we need to distribute binaries at all for Linux users.

dukeeeey:
>Couldn't get the code to compile, had to add #ifdef around "SDL_HINT_WINDOWS_DPI_AWARENESS" in Src/OSD/SDL/Main.cpp

maybe outdated sdl?

PR #226, opened 2025-02-17
Hooking up MIDI output from soundboard

gm-matthew:
Used by Magical Truck Adventure to check region code; ROM in Games.xml is Export version, not Japan.

The patch to enable region change is no longer required and has been removed. A new patch to set the region to Japan has been added but commented out to disable it by default.

I don't really like how I have hardcoded the MIDI FIFO sizes in CheckMIDIStatus(); I think adding functions to SCSP.cpp such as SCSP_MidiInFull() and SCSP_MidiOutEmpty() would be a cleaner way to do it. Thoughts?

### DIFF
--- a/Src/Model3/SoundBoard.cpp
+++ b/Src/Model3/SoundBoard.cpp
@@ -330,11 +330,28 @@ int SCSP68KRunCallback(int numCycles)
  Sound Board Interface
 ******************************************************************************/
 
+UINT8 CSoundBoard::ReadMIDIPort(void)
+{
+	UINT8 result = SCSP_MidiOutR();
+	return result;
+}
+
 void CSoundBoard::WriteMIDIPort(UINT8 data)
 {
 	SCSP_MidiIn(data);
 	if (NULL != DSB)	// DSB receives all commands as well
 		DSB->SendCommand(data);
+}
+
+UINT8 CSoundBoard::CheckMIDIStatus(void)
+{
+	UINT8 status = 0x80;
+	if (SCSP_MidiInFill() < 256)	// MIDI input buffer not full
+		status |= 1;
+	if (SCSP_MidiOutFill() > 0)		// MIDI output buffer not empty
+		status |= 2;
+
+	return status;
 }
 
 #ifdef SUPERMODEL_LOG_AUDIO
--- a/Src/Sound/SCSP.cpp
+++ b/Src/Sound/SCSP.cpp
@@ -1757,8 +1757,8 @@ void SCSP_MidiOutW(BYTE val)
 
 	//printf("68K: MIDI out\n");
 	//DebugLog("Midi Out Buffer push %02X",val);
-	MidiStack[MidiOutW++]=val;
-	MidiOutW&=31;
+	MidiOutStack[MidiOutW++]=val;
+	MidiOutW &= 3;
 	++MidiOutFill;
 
 	if (s_multiThreaded)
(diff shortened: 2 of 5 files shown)


Src/Sound/SCSP.cpp
```diff
@@ -1757,8 +1757,8 @@ void SCSP_MidiOutW(BYTE val)
 
 	//printf("68K: MIDI out\n");
 	//DebugLog("Midi Out Buffer push %02X",val);
-	MidiStack[MidiOutW++]=val;
-	MidiOutW&=31;
+	MidiOutStack[MidiOutW++]=val;
+	MidiOutW &= 3;
```
trzy, line 1761: The buffer is only 4 bytes?

Src/Model3/SoundBoard.cpp
```diff
@@ -330,13 +330,30 @@ int SCSP68KRunCallback(int numCycles)
...
 {
 	SCSP_MidiIn(data);
 	if (NULL != DSB)	// DSB receives all commands as well
 		DSB->SendCommand(data);
 }
 
+UINT8 CSoundBoard::CheckMIDIStatus(void)
+{
+	UINT8 status = 0x80;
+	if (SCSP_MidiInFill() < 256)	// MIDI input buffer not full
```
trzy, line 349: How come 256 is used here but it appears that inside of SCSP.cpp, the FIFOs are only of size 4? It may indeed make sense to have either a function or a static constexpr var that specifies the buffer length.

Src/Sound/SCSP.cpp
gm-matthew: Yes, according to [the official SCSP documentation](https://segaretro.org/images/2/2f/ST-077-R2-052594.pdf).

Src/Model3/SoundBoard.cpp
gm-matthew: The real MIDI input FIFO buffer is only 4 bytes, but it is increased to 256 bytes in Supermodel simply because the soundboard doesn't synchronize with the mainboard in the emulator as often as it does on real hardware. Some games (Sega Rally 2 in particular) write several sound commands in a single frame so the increased FIFO size is a way of making sure they are all received by the soundboard.

As a side note, the MIDI input FIFO buffer was actually originally 128 bytes but I increased it to 256 because Sega Rally 2 was sending so many commands it was overflowing the FIFO buffer (this commit adds a check to prevent overflows).

In SCSP.cpp there is MIDI_STACK_SIZE which is defined as 0x100 (256) but SoundBoard.cpp doesn't have access to this. If I were to create a function within SCSP.cpp such as SCSP_MidiInFull() I could use the existing MIDI_STACK_SIZE; probably a good idea to turn it into a constexpr as well.

Src/Sound/SCSP.cpp
trzy: I see. In that case, shouldn't SoundBoard.cpp use 4 rather than 256? Definitely makes sense to define that constant in SCSP.cpp or SCSP.h. I think your original suggestion of functions that indicate full vs. empty makes the most sense because it doesn't make sense for external components like CSoundBoard to be worrying about FIFO size explicitly.

Src/Model3/SoundBoard.cpp
trzy: Ok, I understand, that makes sense. Ignore what I said in the comment I just posted, then. I think the functions you propose make sense and would be good to include this note about the buffer being larger due to less frequent synchronization.

dukeeeey: What's the status on this? Was there a reason it didn't get merged?

gm-matthew: Still haven't finished refining the code, in particular I wanted to remove the hardcoded FIFO sizes. I will get around to it at some point

PR #229, opened 2025-02-19
Implement "Block Culling" modeword bit of Mercury ASIC

gm-matthew:
Disables 3D rendering and clears screen to white. Only implemented in new 3D renderer for now, though it could easily be added to the legacy 3D renderer if desired.

I added an option to turn this off if desired; set NoWhiteFlash to true in Supermodel.ini or add -no-white-flash to command line.

### DIFF
--- a/Src/Graphics/Legacy3D/Legacy3D.cpp
+++ b/Src/Graphics/Legacy3D/Legacy3D.cpp
@@ -967,6 +967,19 @@ void CLegacy3D::RenderFrame(void)
   if (m_aaTarget) {
       glBindFramebuffer(GL_FRAMEBUFFER, m_aaTarget);			// if we have an AA target draw to it instead of the default back buffer
   }
+
+  if (blockCulling && !m_config["NoWhiteFlash"].ValueAs<bool>())    // block culling disables 3D rendering
+  {
+      // clear screen to white
+      glClearColor(1.0f, 1.0f, 1.0f, 1.0f);
+      glClear(GL_COLOR_BUFFER_BIT);
+
+      if (m_aaTarget) {
+          glBindFramebuffer(GL_FRAMEBUFFER, 0);
+      }
+
+      return;
+  }
   
   // Z buffering (Z buffer is cleared by display list viewport nodes)
   glDepthFunc(GL_LESS);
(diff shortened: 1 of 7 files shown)


Src/OSD/SDL/Main.cpp
```diff
@@ -1612,6 +1613,7 @@ static void Help(void)
   puts("  -legacy3d               Legacy 3D engine (faster but less accurate)");
   puts("  -multi-texture          Use 8 texture maps for decoding (legacy engine)");
   puts("  -no-multi-texture       Decode to single texture (legacy engine) [Default]");
+  puts("  -no-white-flash         Disables white flash when games disable 3D rendering");
```
trzy, line 1616: Why is this option needed? Do games frequently toggle rendering and cause an annoying flash?

Src/Graphics/Legacy3D/Legacy3D.cpp
```diff
@@ -1310,6 +1310,10 @@ void CLegacy3D::SetSunClamp(bool enable)
 {
 }
 
+void CLegacy3D::SetBlockCulling(bool enable)
+{
+}
```
trzy, line 1315: Would be nice to have this implemented. I can look into it later this week on Friday if it's not obvious how to do it in the legacy renderer. Just need to know which games do this.

trzy (commented):
Looks good to me although would be nice to support in legacy renderer. I did not realize this is what the feature did. "Block culling" sounds like it would just toggle whether scene node-level culling is applied or not.

Src/OSD/SDL/Main.cpp
gm-matthew: Jiterdoomer suggested that I make it an option [on Discord:](https://discord.com/channels/609423593063055409/609425438686576682/1180996928780128376)

> Make it toggleable so that people who had epilepsy wont get bothered with white flash feature

If you don't think it's necessary I could always remove the option.

trzy: No, this sounds fine. I didn't realize it manifested itself as a flash. In that case, it really is an accessibility feature. I like the way you've labeled it as "no white flash", as that's a good description for people who would be affected.

Src/Graphics/Legacy3D/Legacy3D.cpp
gm-matthew: Done :-)

trzy (approved):
Awesome! Just tested it with new and legacy engines on lamachin. Will merge.

PR #253, opened 2025-04-25
Add network output, based on MAME's network output system

njz3:
Add TCP server socket 8000 and UDP (broadcast) 8001 for network outputs.

### DIFF
--- a/Src/Network/UDPSend.h
+++ b/Src/Network/UDPSend.h
@@ -1,0 +1,45 @@
+#ifndef _UDPSEND_H_
+#define _UDPSEND_H_
+
+#include "UDPPacket.h"
+#include "WinSockWrap.h"
+#include <string>
+#include <vector>
+#include <thread>
+
+namespace SMUDP
+{
+	class UDPSend
+	{
+	public:
+		UDPSend();
+		~UDPSend();
+
+		bool Send(const char* address, int port, int length, const void *data, int timeout);		// blocking
+		bool SendAsync(const char* address, int port, int length, const void *data, int timeout);	// no blocking call
+
+	private:
+
+		bool WaitForEvent(SOCKET s, HANDLE hEvent, long nEvents, int timeout);
+		bool ProcessReply(SOCKET s, HANDLE event, int timeout);
+		int SendDataPacket(Packet& p, SOCKET s, SOCKADDR_IN& address, HANDLE events);
+		
+		SOCKET m_socket;
+		HANDLE m_event;
+		WinSockWrap m_winsockWrap;
+
+		// async
+		void SendThread();
+
+		HANDLE m_exitEvent;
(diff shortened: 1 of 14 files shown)


Src/Network/UDPSend.h
```diff
@@ -0,0 +1,45 @@
+#ifndef _UDPSEND_H_
+#define _UDPSEND_H_
+
+#include "UDPPacket.h"
+#include "WinSockWrap.h"
+#include <string>
+#include <vector>
+#include <thread>
+
+namespace SMUDP
```
trzy, line 10: Do we need this namespace at all?

Src/OSD/SDL/Main.cpp
```diff
@@ -1582,6 +1583,9 @@ static Util::Config::Node DefaultConfig()
 #endif
   config.Set("Outputs", "none");
   config.Set("DumpTextures", false);
+  config.Set("NetOutputsWithLF", "1");
```
trzy, line 1586: I think we can probably drop the Net prefix here and just group these below "Outputs".

trzy, line 1586: OutputsWithLF should probably be a bool rather than int?

trzy (changes requested):
It's a good feature but substantial rework needed. I don't like the Windows-specific nature of this. I think it's unnecessary: UDP should be done via SDL2_net, not Winsock, and threading should be done with std::thread, not the Windows API.

Also, make sure to add the Supermodel file header to each source file.

dukeeeey:
All the network stuff is copy/paste from much earlier versions of supermodel before the networking code really worked. They shouldn't be needed I don't think for the mame outptut stuff. Tbh not even sure what it is / for.

trzy:
The main thing is that the current outputs system is constrained to Windows but this one could actually be platform-independent. Better to do that.

dukeeeey:
All the network code is what I wrote for spindizzi when we trying to get networking working. It wasn't required for the mame output stuff. In the end i rewrote the network code for SDL.

njz3:
@dukeeeey  That's something I have done a long time ago for my own usage (forcefeedback software), maybe a few years ago), and forgot about it until somebody ask me an update today.
I simply updated the branch up to current master, and commited the files which explains why it is using old code from you.

When I will have spare time I will check your code review. Thank for spending time reviewing it.

dukeeeey:
The only files you need are netoutputs.cpp / h. I am pretty sure the rest isn't needed.

trzy:
Revisiting this. In light of the merged PR (https://github.com/trzy/Supermodel/pull/303), is this PR still needed or should I close it? If I understand correctly, network outputs already exist in Supermodel.

njz3:
Yes the PR can be cancelled and close. Thanks for following up.

PR #258, opened 2025-07-24
fix -config-inputs not registering any inputs

wsxdrfvyhn:


toxieainc:
Seems this was my bad, sorry. Could you please add an additional comment that the parameter passing is intended like this?
(as static analysis warned me, that this could be the wrong way around due to the naming)

wsxdrfvyhn:
done, will look further into it to see if i can come up with an explanation for why exactly this is needed

trzy:
This looked gross to me but I looked at the logic and it made sense:

```
 // See which sources activated to begin with and from here on ignore these (this stops badly calibrated axes that are constantly "active"
  // from preventing the user from exiting read loop)
  ```

The badSources array is being initialized by this call to any "noisy" sources that are initially active. Let me test this tonight and then merge it.

I wonder if it makes sense to make the code more explicit by adding a noisySources vector, and then explicitly copying that to badSources after that initial CheckAllSources() call. It would stop automatic checkers from complaining. Might also add a more detailed comment about what is going on, although I think the existing comment Nik originally makes clear what is happening.

dukeeeey:
Is there any update on this? Unfortunately I don't have any controllers to test with atm

PR #260, opened 2025-08-06
Update README.md

laromicas:
Added needed library for linux compiling.


sgpowers:
It's already in the workflow: 

https://github.com/DirtBagXon/model3emu-code-sinden/blob/main/.github/workflows/main.yml#L51

PR #282, opened 2025-10-19
[DRAFT] libretro core proof-of-concept

jeapostrophe:
This PR has a basic version of a libretro core.

There are some obviously ugly things that would be a lot more beautiful with some (hopefully small) changes to Supermodel, but it performs very well on OS X and Windows with 'Spikeout: Final Edition'.

# Things I wish were a little different about Supermodel
- Main --- I copied a bunch of code from Src/OSD/SDL/Main.cpp, which I would otherwise call if those functions were public. `CreateGLScreen` is almost identical, although I modified my copy to hide the window.
- Configuration --- I would have liked to have an easy enumeration of all the config variables that Supermodel looks for and then I would call them through the libretro environment callback. Instead, I've just hard-coded it.
- SaveStates --- It would be very nice if there were code not only in SDL/Main that dealt with save states and it would be even better if they could work without files. My implementation writes a temporary file and then reads it. It would also be nice to be able to predict the size of the state without writing it, but I don't know if that is possible.
- Audio --- The existing Audio system relies on implementations providing C functions, rather than using an object interface. If there were an object interface, it would be a lot cleaner. I think I understand why it does it the way it does because it is hard to get the audio callback pointer deep into the bowels of the sound board code. Maybe a compromise would be to abstract the Audio OSD interface with an SDL version and a version like this? I worked around this with a #define

# Things I'm not sure how you'd prefer things to be implemented
- Building --- I've made it so there's a new build path that supports building the binary, but it actually doesn't work. Either the build system right now doesn't have a good way of giving different #define; although if the Audio issue is resolved, it wouldn't be necessary.

# Hacks I made that are possible to fix but I just haven't done
- Input --- My InputSystem implementation is an okay way of working with libretro, but a "better" way would be to reflect the different input types with libretro's input descriptor system.
- OpenGL --- I am hacking by opening up a hidden SDL window and having it do rendering, rather than asking libretro for a context. This works well on desktop, but presumably would fail on mobile. I made a fake pixel format (1337) which is RGBA_UpsideDown, but it would be better to use XRGB8888 and do the transformation inside the core. My fake format is fine for my testing because I have a testing front-end that uses RGBA anyways.
- NVRAM --- I've ignored this, but libretro provides a way to look at a save directory.

I'm posting this to get feedback on the above points and advertise in case anyone else is interested in working on this.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,56 +1,121 @@
-name: Build and Release
+name: Release Supermodel Libretro
 
 on:
   push:
-    branches: [ master ]  # Auto-build on master
-  workflow_dispatch:      # Manual triggering
+    tags:
+      - "v*"
+  workflow_dispatch:
     inputs:
-      opt:
-        description: "Optimization flags"
+      tag_name:
+        description: "Release tag to create or publish (for example v0.3a-1)"
+        required: false
         type: string
-        default: "-O3 -g"
+        default: ""
+      release_name:
+        description: "Release title"
+        required: false
+        type: string
+        default: ""
+      notes:
+        description: "Extra release notes to prepend"
+        required: false
+        type: string
+        default: ""
+      prerelease:
+        description: "Mark as prerelease"
+        required: false
+        type: boolean
+        default: true
 
 jobs:
   build:
     uses: ./.github/workflows/build.yml
     with:
-      opt: ${{ github.event.inputs.opt || '-O3 -g' }}
-  
+      opt: "-O3 -g"
+
   release:
     needs: build
-    if: github.ref == 'refs/heads/master'
     runs-on: ubuntu-latest
     permissions:
       contents: write
     steps:
-      - uses: actions/checkout@v4
-      
-      - name: Get version
-        id: version
-        run: |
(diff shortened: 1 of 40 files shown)


jeapostrophe:
FWIW, I play Spikeout with this almost every day and haven't had to make any changes. I haven't tested any other games. I expect that gun games will be the biggest problem.

SRC267:
> FWIW, I play Spikeout with this almost every day and haven't had to make any changes. I haven't tested any other games. I expect that gun games will be the biggest problem. 

That's probably the only game I'd play, is the core available for android too? I'd like to test.

trzy:
Need to look into this over the holidays!

SRC267:
> > Is the core available for android too? I'd like to test.
> 
> @SRC267 I have not done anything to port to new platforms. I have tested on macOS and Windows and I use Windows regularly. The way I did the core is not very conducive to a mobile port because of the hidden window creation and framebuffer read.

Okay, if you have the windows core, I can try it on my PC. I have a few PCs, I'll try it on my handheld pc.

jeapostrophe:
I just merged a big update that @mixerjdp did to the PR that makes it a lot better and removes most of my hacks. He did a killer job and I think it is so much better now.

PR #289, opened 2025-11-10
Improve the microtexture blending factor

gm-matthew:
It's actually thanks to implementing microtextures in the MAME Model 2 driver that I was able to figure out the proper blending factor, and I am now very confident that it is the same for Model 3 :-)


Garagefreek:
Nice one.

PR #308, opened 2026-03-05
SDL simulated vibration feature added

pizzzza19:
Using SDL_GameController Rumble API commands to simulate wheel board vibration for game pads.
Tested with Xbox one windows controller.
Tested with Daytona USA 2 and Scud Race.
For dayto2pe, you can feel the haptic difference between each car.

Read from codes, Wheel board has 4 functions
* Centering
* Friction
* ConstantForce (FFB torque)
* Vibrate

SDL also has FFB feature and it can be possible to simulate them too. (Future TODO list)
Only vibration can simulate with game pads for now.

All of signals came from main machine boards, so we can path-through the value directly to SDL for skipping board EPROMs.

### DIFF
--- a/Src/Model3/DriveBoard/WheelBoard.cpp
+++ b/Src/Model3/DriveBoard/WheelBoard.cpp
@@ -23,33 +23,211 @@
 /*
  * WheelBoard.cpp
  *
- * Implementation of the CWheelBoard class: drive board (force feedback for wheel)
- * emulation.
+ * This code is enhanced with Claude by ANthropic. 
+ * HLE (High Level Emulation) replacement for the drive board Z80 ROM.
+ * Force feedback commands from the PPC are decoded directly here and
+ * forwarded to SDL Haptic — no Z80 ROM required.
  *
- * NOTE: Using SDL haptic to simulate vibration. Drive board ROMs are no longer required. FFB function still ROMs required.
- * epr-19338a: Scud race / plus
- * epr-19338a: Dart devils
- * epr-20985: Daytona USA 2 PE / BOTE
- * epr-20985: Sega Rally2
+ * Supported games:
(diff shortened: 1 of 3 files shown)


pizzzza19:
Many things changed as previous SDL haptic has problems.
Here is the new codes which added

* Windows and Apple Silicon Mac compatible
* Default disabled which does not affects to users already have original driveboard rom.
  To activate HLE vibration without 64KB ROM, open supermodel.ini and enable below:
     ForceFeedback = 1
     DriveBoardHLE = 1
  (Prepare dummy 64KB rom (FF or Zero filled) inside dayto2pe.zip otherwise boot fail.)
* Tested with DAYTONA2 and SCUD RACE. This will also work on SEGA RALLY 2 but not tested yet as i don't have ROM.

trzy:
This is not acceptable as-is. It completely violates the architectural principles of the project. SDL input code is considered platform-dependent because we support multiple input systems, hence why there is an Src/OSD/SDL/ folder with an SDLInputSystem.cpp file that already implements force feedback. This PR plops all kinds of redundant SDL input code into what is supposed to be platform-agnostic emulation code.

If you want to implement a drive board simulation, please understand how the current non-simulated drive board code works and how it sends force feedback commands to the platform-specific input system. Your code should work with all supported force feedback systems if implemented correctly and there should be absolutely NO platform-specific code, no extra initialization in Main.cpp, etc.

I don't mind if it's done with AI but please clean up AI slop comments and remove any advertisements that Claude inserts.

pizzzza19:
Debugging with AI audit so i just left the comments.
There was absolutely no intention to insert advertisements. I removed those comments from everywhere.

This code works ok windows and mac but if SDL still platform dependent style, need to start over again.
Someone may improve it but i will stay in my private fork as lack of skilling now.

Thank you very much your feedback.

PR #313, opened 2026-05-13
Make shifter behavior work better with shifter controllers

AnthonyRyuki:
I have a Logitech G29 wheel with a shifter at home, and while the existing behavior makes the use of buttons for shifting more user-friendly, it doesn't exactly replicate the actual Sega shifter, which would be a better option with a shifter controller such as my Logitech one.
I do, however, understand that not everyone has a shifter controller at home, so maybe it would be better to implement a new variable that toggles between the original behavior and this one.

### DIFF
--- a/Src/Inputs/InputTypes.cpp
+++ b/Src/Inputs/InputTypes.cpp
@@ -164,8 +164,8 @@ void CGearShift4Input::Poll()
 	prevValue = value;
 
 	// Gears (values 1-4) are set by pressing a button (lower gears have priority) and "stick" until a shift to another gear is made.
-	// Neutral is selected by pressing the neutral gear button. It means all gears are released (value 0).
-	if		(m_shiftNInput->Pressed()) value = 0;
+	// Neutral is selected by releasing all gear inputs (value 0).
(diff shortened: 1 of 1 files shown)


AnthonyRyuki:
Well, I tried to implement a new variable to toggle between the two behaviors, but my programming skills are too poor for that. I'm not going to apply any further changes to this PR.

trzy:
I don't quite understand the change. The original logic was implemented by Nik, who I believe had a G29 or something equivalent. How does the Logitech shifter behave? My understanding is that the neutral position was its own "button", in which case the existing code should work.  Please give me as much detail as possible.

AnthonyRyuki:
With the G29, a "button" on the shifter is only pressed while a gear is inserted, and when the shifter is neutral, no input is registered. From what I've heard, there are Logitech wheels that use the logic that you mentioned (such as the G923), but it's not the case with the G29. So with the original logic, when the shifter is neutral, the game is still considering the gear that was previously inserted to be still in, and the neutral position is never registered.

AnthonyRyuki:
Well, I've just found a solution for this that does not require modifying the code. Just assigning InputGearShiftN to something like "!JOY1_BUTTON1+!JOY1_BUTTON2+!JOY1_BUTTON3+!JOY1_BUTTON4" gets the job done. I'm going to close this silly PR.

PR #318, opened 2026-05-31
Lost World gun detection

ToBul:
When Supermodel loads a merged ROM it ignores the parent.
This causes a problem with The Lost world child ROM as Supermodel isn't looking for/ recognising it.

Can't say I understand the GameLoader code but comments at line 718 onwards seem to explain the design choice,
https://github.com/trzy/Supermodel/blob/77d28eec84a5ededaa23bb31bc74db631cbe5530/Src/GameLoader.cpp#L718

BTW
This is also related to why people have been changing their games.xml from analog_gun1 to gun1 for the last seven years, as that trick bypasses this bug.

It also explains why people think they have a bad Lost World ROM, leading them to download a different ROM from another site. The ROMs often aren't bad, they are just merged.

### DIFF
--- a/Src/Model3/Model3.cpp
+++ b/Src/Model3/Model3.cpp
@@ -608,7 +608,7 @@ UINT8 CModel3::ReadInputs(unsigned reg)
   	  // probably to allow users to define inverted controls for this game only,
   	  // which means the input system must support loading per-game config (not
   	  // all analog_gun games require axis inversion to be playable).
-	  if (m_game.name == "lostwsga" || m_game.name == "lostwsgo")
+	  if (m_game.name == "lostwsga" || m_game.name == "lostwsgp")
 	  { // to do, not a string compare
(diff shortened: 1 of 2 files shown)


ToBul:
Btw, it makes no sense to me to ignore the parent because for a child to work the parent has to be present right? I think most users would expect Supermodel to default to the parent game, principle of least surprise and all that.

trzy:
Yeah I agree with ToBul and I think this needs to be rethought. I’m not exactly up to speed on the specifics here but Supermodel *does* try to load the parent ROM set by assuming the zip file is named the same way.

Shouldn’t the appropriate gun type be determined from the child set? I’m not a big fan of this hack that exists to detect lightgun games by name; that’s what the XML is for.

If we need to revamp ROM loading (and I will continue to insist games are auto detected based on CRC), we should just do that instead.

ToBul:
I guess the best option is to just change the games.xml to gun1 and remove hacky code, it did look like an ugly add on.
I will close this, as not really needed.

PR #323, opened 2026-07-12
Always use MPC105 for step 1.x and MPC106 for step 2.x

gm-matthew:
Some step 1.x games try to detect MPC106 first by accessing it via 0xFEC00CF8 and 0xFEE00CFC but then fail because the device ID is incorrect; we need to make sure MPC105 is only accessible via 0xF0800CF8 and 0xF0C00CFC

Remove <pci_bridge> and <real3d_pci_id> tags from Games.xml as they are no longer needed

### DIFF
--- a/Config/Games.xml
+++ b/Config/Games.xml
@@ -19,7 +19,6 @@
     <hardware>
       <platform>Sega Model 3</platform>
       <stepping>1.0</stepping>
-      <pci_bridge>MPC106</pci_bridge>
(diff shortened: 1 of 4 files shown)


trzy (approved):
Thank you!

gm-matthew:
@trzy You'll need to merge it as I don't have write access yet 😉

trzy:
Oops, done!

Issue #278, opened 2025-10-10
Add the ability to change the path of roms in the GUI


dirctbe:
Hi, would it be possible, now that Supermodel has a GUI, to add the option to choose the path where the emulator will retrieve ROMs to the settings?

Sometimes, for organizational reasons, people don't keep the ROM folder in the same folder as the emulator.

cerroverb:
yes please this would be such a useful feature to have. i'm literally trying to figure out how to change it from the folder demands to the folder where the rest of my ROMs are with emulation station and emudeck.

KAI-SHMUPER:
Any idea if this was implemented after all this time? Because I still dont see the option to change the path even on the latest build's ini.

PR #331, opened 2026-08-06
Add new config key RomsDirectory to Supermodel.ini

hywelandrews:
**Rational**: When using Supermodel from it's GUI roms are currently searched only in a child ROMs/ folder. This is a problem in particular when Supermodel is installed from a system package manager (homebrew, pkg, apt etc). This change allows users to configure any path to be searched when launching roms from the GUI. It is not provided as a command line option, only as a key in Supermodel.ini.

Thanks

### DIFF
--- a/Src/OSD/SDL/Gui.cpp
+++ b/Src/OSD/SDL/Gui.cpp
@@ -778,13 +778,14 @@ static void GUI(const ImGuiIO& io, Util::Config::Node& config, const std::map<st
     ImGui_ImplOpenGL3_RenderDrawData(ImGui::GetDrawData());
 }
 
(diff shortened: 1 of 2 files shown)


trzy:
I need to take a look at this because I thought we had a mechanism for locating directories according to (obtuse) UNIX conventions but that might’ve been only for the config file.

one nit: call it “directory”, not folder, to be consistent with the rest of the app.

hywelandrews:
Sure, I've updated the config key and internal variables to [Rr]omsDirectory. I had a look at the OSD/FileSystem implementations when submitting the FreeBSD port, and all platforms did just have to handle config paths but I've only been able to test this on MacOSX and UNIX. Would also partially resolve #278

PR #332, opened 2026-08-10
Adds OpenGL ES 3.2 & Wayland support

dmanlfc:
- This improves compatibility across Linux platforms
- Allows lower-end Arm devices to use Supermodel
- Avoids using restricted OpenGL drivers and XWayland
- Fix initial rendering, especially when set to Fullscreen.


dmanlfc:
Updated to fall back to OpenGL ES 3.1 also

dukeeeey:
This is nice work but the question for us is do we want to support long term essentially a new rendering backend. I know desktop opengl and es version are pretty close but they aren't identical. Also you've patched imgui itself, which maybe problematic if we want to update these files.

trzy:
I agree — who does an ES backend benefit (other than maybe people trying to make for-profit ports)? 

I’d like to remove my old renderer at some point as I don’t think it’s particularly useful anymore.

dmanlfc:
Most newer Arm platforms don't provide OpenGL desktop drivers only OpenGL ES hence the requirement. Those that do, sometimes have a lower shader version than the one now supported with Supermodel. i.e the RK3576 chipsets.

trzy:
I imagine these platforms aren’t good for gaming in general. What is the use case? Building MAMECabs that are less expensive than PC based ones?

dmanlfc:
> I imagine these platforms aren’t good for gaming in general. What is the use case? Building MAMECabs that are less expensive than PC based ones?

Correct but Supermodel runs well on lower end devices just fine with the right driver implementation and the influx of Arm based handhelds has this requirement. Driver development and focus is in mobility, so OpenGL ES & Vulkan rather than OpenGL Desktop.

jeapostrophe:
FWIW, I have recently had Fable write a Vulkan port of New3D. I haven't tested every game (I just play Spikeout) but it looks perfect to me and the changes a pretty minor. I thought about using GL ES and decided to go with Vulkan for the forward stability of it. I'll update my fork shortly to include it. [Right now, Vulkan is only exposed on the libretro path.]

dmanlfc:
I already have Vulkan in my fork also.

PR #336, opened 2026-09-20
GUI preserves per-game config settings

trzy:
Currently, the GUI loads and writes only the Global (root-level) config settings, which means that any per-game settings are thrown away when the configuration is saved from within the GUI. This PR addresses the issue by passing the *entire* configuration (including per-game settings) to the GUI and making it update only the top-level settings by copying them in.

I have also modified the configuration file header comment. It now includes a timestamp of when the file was last written by Supermodel and where it was written from (saved via GUI, updated via input configuration, or default configuration).

Note: this does *not* add per-game settings to the GUI. It merely ensures the GUI preserves any per-game config settings that have been added by manually editing Supermodel.ini.


dukeeeey:
The per gui stuff shouldn't be that hard to add to the gui. I think how i wanted it to work was simply simply to pass a config which would be written to by the gui. Then essentially do a diff against the global to get a list of only what was changed and store that. Where i started to get tripped up was simply the nodes themselves, they seem to have a linked list then child nodes etc. The more i looked the crazier it got lol.

trzy:
Should we merge this change or forego it? 

Re: how the Node stuff works:

Util::Config::Node is a nested structure, as you know, that maps most naturally onto XML semantics. INI files are basically flat but have sections. So the config *file* has the config tree structured a very particular way:

Top-level is implicitly "global" and contains a list of key, value pairs (that's what a Node is, IIRC). They should have no children. You can write a Supermodel.ini file with just stuff like "WideScreen=true" and no section headers. These are all implicitly under [Global].

But INI files can have sections (e.g., [scudau]). These are nodes that have children (I think the value is just ignored in this case). So when we write an INI file, the top-level global settings get dumped under [Global], and then anything that has child nodes gets interpreted as a section header, and its direct children are written out as settings beneath it.

So the following INI file:

```
WideScreen=true
SoundVolume=50

[scudau]
WideScreen=false
SoundVolume=100
```

Would have a Util::Config::Node tree like so:

```
Global=
  WideScreen=true
  SoundVolume=50
  scudau=
    WideScreen=false
    SoundVolume=100
```

Does that make sense?

So when you have e.g. `Util::Node::Config fileConfig`, that node itself has its key either empty or explicitly set to "Global" (either way, the INI file writer treats it as Global), and then you ignore its value but iterate its children. Those are the settings you are parsing now.

The per-game settings appear as further descendants.

dukeeeey:
Yeah that makes sense, I think the part that actually tripped me up was the linked lists? Not even sure why they are used. Im not at my pc but haven't really looked at the PR but I'm sure it's fine.

